PR #214, opened 2024-05-10
Complex navigator first version

jmedinaebi:
All the changes for the 1st releasable version of the Complex Navigator.


src/app/complex/complex-results/complex-filter/complex-filter.component.ts
```diff
@@ -1,6 +1,8 @@
 import {Component, EventEmitter, Input, OnInit, Output} from '@angular/core';
 import {Facets} from '../../shared/model/complex-results/facets.model';
 import {AnalyticsService} from '../../../shared/google-analytics/service/analytics.service';
+import {Interactor} from '../../shared/model/complex-results/interactor.model';
```
jmedinaebi: Are these import needed?

src/app/complex/complex-results/complex-navigator-sorting/complex-navigator-sorting.component.html
```diff
@@ -0,0 +1,49 @@
+<div class="filters">
+
+  <div class="sortingInteractors">
```
jmedinaebi: We should either split this into 2 components, or rename the component.

It's called complex-navigator-sorting at the moment, and it contains 2 things, the menu for sorting, and then the menu for the view; so the name is slightly wrong now.

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-subcomponent-interactor/table-subcomponent-interactor.component.ts
```diff
@@ -0,0 +1,163 @@
...
+  displayBottomLineClass: string;
+  interactorInComplex = false;
+  interactorStoichiometryText: string;
+  interactorStoichiometryValue: string;
+  interactorInSubComplex = false;
+  subComplexInteractorStoichiometryText: string;
+  subComplexInteractorStoichiometryValue: string;
+
+
+  // findInteractorInComplex = findInteractorInComplex;
```
jmedinaebi: Commented out lines, please delete.

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-subcomponent-interactor/table-subcomponent-interactor.component.html
```diff
@@ -0,0 +1,12 @@
+<div [ngClass]="displayTopLineClass"></div>
+<cp-table-interactor-stoichiometry
+  *ngIf="interactorInComplex"
+  [stoichiometryText]="interactorStoichiometryText"
```
jmedinaebi: Not completely necessary, but the same thing we did in this component, having the values in variables set in ngOnChangesn and not calling methods in ngIf; we could do in other components too.

src/app/complex/complex-results/complex-results.component.ts
```diff
@@ -194,4 +221,40 @@ export class ComplexResultsComponent implements OnInit, AfterViewInit {
...
+    this.DisplayType = this.LIST_VIEW;
+    this.reloadPage();
+  }
+
+  setComplexNavigatorView() {
+    this.DisplayType = this.COMPLEX_NAVIGATOR_VIEW;
+    this.reloadPage();
+  }
+
+  setFirstDisplayType(): void {
```
jmedinaebi, line 266: `window.location.href` is not the proper way to redirect to the complex details page. Also, you're using localhost, which worked for testing locally, but it won't remote.

Try replacing it with
```
  setFirstDisplayType(): void {
    if (this._complexSearch.elements.length === 1) {
      const complexId = this._complexSearch.elements[0].complexAC;
      if (!!complexId) {
        // For some reason this is needed so the navigate call works
        this.router.routeReuseStrategy.shouldReuseRoute = function () {
          return false;
        };
        this.router.navigate(['/complex', complexId]);
      }
    }

    if (!this.DisplayType) {
      // Currently the list view is the default, as we are just launching the navigator view
      // Later on we can change the default view to be the list or navigator view based on number of results
      this.setListView();
    }
  }
```

I also changed the logic of the display type to have list be default for now, so when we first launch it we keep list as the default view. Maybe for the UX testing, we can have the navigator view as default, and then switch it back.

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-name/table-interactor-name.component.html
```diff
@@ -0,0 +1,36 @@
...
+    <div class="name" title="More information about {{interactorId}}" *ngIf="!IDDisplay">
+      <a *ngIf="!IDDisplay" href="{{identifierLink}}" target="_blank"
+         class="externalLinkContainer">
+        {{ interactorName }}
+      </a>
+    </div>
+
+    <a [routerLink]="['/complex/search']"
+       [queryParams]="{query: interactorId, page: 1}"
+       [fragment]="'view_complex_navigator'"
```
jmedinaebi: Instead of `[fragment]="'view_complex_navigator'"`, we can use `[preserveFragment]="true"` to keep the same fragment we currently have in the URL. That way, if we rename it, then it does not matter.

jmedinaebi:
One more thing I noticed, the sorting menu goes behind the complex navigator table

Screenshot 2024-05-10 at 09 47 15

It does not happen with the view menu

Screenshot 2024-05-10 at 09 47 19

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.css
```diff
@@ -118,3 +118,7 @@ a, a:visited {
   white-space: nowrap;
   overflow: hidden;
 }
+
+.container {
+  font-size: 14px;
```
jmedinaebi, line 137: What's this change for?

src/app/complex/complex-results/complex-results.module.ts
```diff
@@ -27,7 +27,7 @@ import {
 import {
   TableSubcomponentInteractorComponent
 } from './complex-navigator/table-structure/table-interactor-column/table-subcomponent-interactor/table-subcomponent-interactor.component';
-import {ComplexNavigatorSortingComponent} from './complex-navigator-sorting/complex-navigator-sorting.component';
+import {ComplexNavigatorSortingComponent} from './complex-navigator-buttons/complex-navigator-sorting.component';
```
jmedinaebi: You rename the component selector `cp-complex-navigator-buttons` and rename the folder containing the component `complex-navigator-buttons`; but the actual component remains with the old name.

Files inside ``complex-navigator-buttons`:
- complex-navigator-sorting.component.ts
- complex-navigator-sorting.component.spec.ts
- complex-navigator-sorting.component.css
- complex-navigator-sorting.component.html

And the component is still called `ComplexNavigatorSortingComponent`.

Nothing important but we should rename all of this for consistency.

src/app/complex/complex-results/complex-navigator/table-structure/table-header/table-header.component.html
```diff
@@ -0,0 +1,66 @@
...
+                   title="{{ complex.complexName }} - {{ complex.complexAC}}"></i>
+                {{ complex.complexName }}
+              </div>
+            </span>
+          </div>
+        </a>
+      </th>
+      <div class="spaceHolder"></div>
+    </tr>
+    <tr [style.--cols]="complexes.length" [style.--intHeader]="isInteractorSortingSet() ? '156px':''">
```
jmedinaebi: Should we have `*ngIf="organismIconDisplay"` here? Not sure about it, but if we add it, we then also need to add it in line 18.

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.css
susiehgt: It is the size of the expandable button which was bigger than the rest

src/app/shared/notification/service/notification.service.ts
```diff
@@ -45,4 +51,12 @@ export class NotificationService {
   public onFeatureNotAvailableYet() {
     this.addHintNotification('This feature is not available yet. But it is coming soon! :-)');
   }
+
+  public rateComplexNavigator(): void {
+    this.addRatingNotification('<div><input class="input-small" value="textbox"/><a href="' + environment.complex_portal_support_url + '">Tell us about your experience with the Complex Navigator!</a></div>');
```
jmedinaebi: Add `target="_blank"` to the `<a>` element `<a target="_blank" href=...`, so it opens a new tab.

src/app/shared/notification/service/notification.service.ts
```diff
@@ -45,4 +51,12 @@ export class NotificationService {
...
+  }
+
+  public complexNavigatorAnnouncement() {
```
jmedinaebi: At the moment the pop up displays at the very top, and it's easy to miss. We can configure it to display lower so it displays on the white part of the screen.

1. In `styles.css` add the following class:
```
.toast-right-under-header {
  right: 10px;
  top: 225px;
}
```

2. Update this method to use that position class:
```
  public complexNavigatorAnnouncement() {
    this.addAnnouncementNotification(
      'Try out our new display: the Complex Navigator!',
      {
        positionClass: 'toast-right-under-header'
      });
  }
```

Give it a try, and you can also use it for the `rateComplexNavigator` if you want.

jmedinaebi: There are more options, like `disableTimeOut` and `closeButton`. So we can disable the timeout and force users to have to close the pop up.

src/app/shared/notification/service/notification.service.ts
```diff
@@ -53,10 +55,26 @@ export class NotificationService {
...
+    `;
+    this.addRatingNotification(ratingHtml);
   }
 
   public complexNavigatorAnnouncement() {
-    this.addAnnouncementNotification('Try out our new display: the Complex Navigator!', this.toastrService.toastrConfig.timeOut);
+    this.addAnnouncementNotification(
+      'Try out our new display: the Complex Navigator!',
+      {
+        positionClass: 'toast-right-under-header'
```
jmedinaebi: So it doesn't close automatically and people keep seeing it, add the following options:
```
        positionClass: 'toast-right-under-header',
        disableTimeOut: true,
        closeButton: true
```

src/app/shared/notification/service/notification.service.ts
```diff
@@ -53,10 +55,26 @@ export class NotificationService {
...
+      <span class="star" [ngClass]="{'selected': selectedRating >= 1}" (click)="selectRating(1)"> &#9733; </span>
+      <span class="star" [ngClass]="{'selected': selectedRating >= 2}" (click)="selectRating(2)"> &#9733; </span>
+      <span class="star" [ngClass]="{'selected': selectedRating >= 3}" (click)="selectRating(3)"> &#9733; </span>
+      <span class="star" [ngClass]="{'selected': selectedRating >= 4}" (click)="selectRating(4)"> &#9733; </span>
+      <span class="star" [ngClass]="{'selected': selectedRating >= 5}" (click)="selectRating(5)"> &#9733; </span>
+    </div>
+    <input type="submit" name="submitButton" value="Submit Rating" class="button" (click)="submitRating()">
+  </div>
+    `;
+    this.addRatingNotification(ratingHtml);
```
jmedinaebi: I've tried but I haven't managed to make the buttons inside the toast clickable. I even used `toastComponent` and created a new component but no luck. I did a quick search and it may be that this type of toast messages don't work with clickable buttons in it.

src/app/complex/complex-results/complex-navigator/complex-navigator-rating/complex-navigator-rating.component.ts
```diff
@@ -0,0 +1,33 @@
...
+  constructor(private analyticsService: AnalyticsService) {
+  }
+
+  selectRating(rating: number) {
+    this.selectedRating = rating;
+    console.log(`Selected rating: ${rating}`);
+  }
+
+  submitRating() {
+    if (this.selectedRating !== null) {
```
jmedinaebi: Replace with
```
if (!!this.selectedRating) {
```

Initially `selectedRating` is undefined, not null (typescript differentiate between these two values), so because selectedRating is not null, it goes in this if and tries to do selectedRating.toString(), which results in 
```
ERROR TypeError: Cannot read properties of undefined (reading 'toString')
    at ComplexNavigatorRatingComponent.submitRating (complex-navigator-rating.component.ts:23:70)
```

`!!this.selectedRating` is equivalent to write `this.selectedRating != null and this.selectedRating != undefined`.

The only thing that can be tricky with !! is that it is more than just null and undefined, 0 may be also one of the values considered. But, because we don't have a 0 rating value, then it's ok.

src/app/shared/notification/service/notification.service.ts
```diff
@@ -25,6 +25,14 @@ export class NotificationService {
     this.toastrService.info(announcementNotification, 'Just to let you know!', options);
   }
 
+  private addComplexNavigatorAnnouncementNotification(announcement: string): void {
+    this.toastrService.toastrConfig.disableTimeOut = true;
```
jmedinaebi: I think when we do `this.toastrService.toastrConfig.disableTimeOut = true;`, we're changing the config for all toasts, not just for this announcement, so if any other toast gets displayed, it will also not have the timeout.

The way to do it only for this is
```
    return this.toastrService.info(announcement, null, {
      disableTimeOut: true,
      closeButton: true,
      progressBar: false,
      positionClass: 'toast-right-under-header'
    });
```
2nd argument is null as it is the title, which we do not set.

src/app/shared/notification/service/notification.service.ts
```diff
@@ -45,4 +53,14 @@ export class NotificationService {
...
     this.addHintNotification('This feature is not available yet. But it is coming soon! :-)');
   }
+
+  public complexNavigatorAnnouncement() {
+    this.addComplexNavigatorAnnouncementNotification(
+      'Try out our new display: \n ' +
+      'the Complex Navigator!');
+  }
+
+  public closeAnnouncement(): void {
```
jmedinaebi: As discussed, updated `addComplexNavigatorAnnouncementNotification` to return the result the toast.info call 
```
return this.toastrService.info(...);
```

Then, update `complexNavigatorAnnouncement` to also return
```
return this.addComplexNavigatorAnnouncementNotification(...)
```

Then, in complex-results.component.ts, store the returned value
```
this.toast = this.notificationService.complexNavigatorAnnouncement();
```

Update `closeAnnouncement` to take the toast id
```
  public closeAnnouncement(toastId: number): void {
    this.toastrService.clear(toastId);
  }
```

And finally, update the call to it
```
setComplexNavigatorView() {
    this.DisplayType = this.COMPLEX_NAVIGATOR_VIEW;
    this.reloadPage();
    if (!!this.toast) {
      this.notificationService.closeAnnouncement(this.toast.toastId);
      this.toast = null;
    }
  }
```

src/app/complex/complex-results/complex-navigator/complex-navigator-rating/complex-navigator-rating.component.ts
```diff
@@ -0,0 +1,33 @@
...
+      this.analyticsService.rateComplexNavigator(this.selectedRating.toString());
+      console.log(`Submitting rating: ${this.selectedRating}`);
```
jmedinaebi: We need to close this popup when the rating is submitted.

src/app/complex/complex-results/complex-navigator-buttons/complex-navigator-buttons.component.ts
```diff
@@ -0,0 +1,74 @@
...
+        } else if (typeOfSorting === 'Organism') {
+          this.organismIconDisplay = false;
+        }
+    }
+    this.changesEmitter();
+  }
+
+  DisplayingOrganism() {
+    this.organismIconDisplay = !this.organismIconDisplay;
+    this.changesEmitter();
```
jmedinaebi: If `organismIconDisplay` is set to true, we need to set detailed display to true too. It may already be true, but it doesn't hurt setting it anyway.

```
    this.organismIconDisplay = !this.organismIconDisplay;
    if (this.organismIconDisplay) {
        this.IDDisplay = true;
    }
    this.changesEmitter();
```

Same change needed in `DisplayingType` and `DisplayingID` with the relevant variables.

This is to fix the issue of, being in compact display and clicking on one of the detailed checkboxes.

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.html
```diff
@@ -0,0 +1,97 @@
...
+          </td>
+        </ng-container>
+        <div class="spaceHolder" *ngIf="complexes.length>8"></div>
+      </tr>
+      <!-- Expandable menu for subcomplexes -->
+      <ng-container *ngIf="interactor.expanded">
+        <ng-container *ngIf="!!interactor.subComponents">
+          <tr *ngFor="let el of interactor.subComponents; let j=index"
+              class="expandedRows"
+              [ngClass]="{'firstExpandedRow': j === 0, 'lastExpandedRow': j === interactor.subComponents.length - 1}">
```
jmedinaebi, line 56: The issue with the cells not having the right width with a low number of complexes is in this line. I'm still looking into it, but this is definitely the line.

jmedinaebi: I think I got it.

In `table-interactor-column.component.css`, for class `interactorsColumn`, get rid of `width: 130px;` (line 16); and for class `intStoich`, get rid of `width: 70px;` (line 32).

susiehgt: I think I've put it this way because of the interactors header which was not sizing the right way but I will check that tomorrow because it was working well on the previous versions

jmedinaebi:
I've tested how to handle pagination for list view and complex view separately and it's easier than I thought. We don't need to move the paginator to the list and complex navigator components. We just need to rely on setters and getters to set or get the right value.

1. Delete the current variables to store pagination values
    ```
    private _currentPageIndex: number;
    private _lastPageIndex: number;
    private _pageSize = 15;
    ```

2. Add new variables, having separate ones for list and navigator view
    ```
      private _listPageSize = 15; // This is where we set the size of the pages for list view
      private _navigatorPageSize = 30; // This is where we set the size of the pages for navigator view
      private _listCurrentPage: number;
      private _navigatorCurrentPage: number;
      private _listLastPageIndex;
      private _navigatorLastPageIndex;
    ```

3. Update setters and getters for `currentPageIndex`, `lastPageIndex` and `pageSize`. They would set or get the right variable based on the display type currently set.
    ```
      get currentPageIndex(): number {
        if (this.DisplayType === this.COMPLEX_NAVIGATOR_VIEW) {
          return this._navigatorCurrentPage;
        } else {
          return this._listCurrentPage;
        }
      }
    
      set currentPageIndex(value: number) {
        if (this.DisplayType === this.COMPLEX_NAVIGATOR_VIEW) {
          this._navigatorCurrentPage = value;
        } else {
          this._listCurrentPage = value;
        }
      }
    
      get lastPageIndex(): number {
        if (this.DisplayType === this.COMPLEX_NAVIGATOR_VIEW) {
          return this._navigatorLastPageIndex;
        } else {
          return this._listLastPageIndex;
        }
      }
    
      set lastPageIndex(value: number) {
        if (this.DisplayType === this.COMPLEX_NAVIGATOR_VIEW) {
          this._navigatorLastPageIndex = value;
        } else {
          this._listLastPageIndex = value;
        }
      }
    
      get pageSize(): number {
        if (this.DisplayType === this.COMPLEX_NAVIGATOR_VIEW) {
          return this._navigatorPageSize;
        } else {
          return this._listPageSize;
        }
      }
    ```

4. Update any part of the code using the variables we deleted in step 1 directly instead of the setters or getters. No changes are needed in the HTML, but some are needed in the ts file.

    4.1. In `ngOnInit` method, 5th line inside the `this.route.queryParams.subscribe` body.
    ```
    this.currentPageIndex = queryParams['page'] ? Number(queryParams['page']) : 1;
    ```
    4.2. In `reloadPage` method, 3rd line of the method
    ```
    queryParams['page'] = this.currentPageIndex;
    ```
    Other parts of the code, like setting the last page, already use the setters and getters, so no more changes are needed.

jmedinaebi:
Regarding my latest comment about the pagination changes, there is one issue. When the page is reloaded, the body in `this.route.queryParams.subscribe` runs before the body in `this.route.fragment.subscribe`. This means that we try to do the search, and set pagination variables, before we set the display.

We can fix it but having one subscribe inside the other. Instead of
```
this.route.queryParams.subscribe(queryParams => {
    ...
});

this.route.fragment.subscribe(fragment => {
    ...
});
```
As we currently have it, we can do
```
this.route.fragment.subscribe(fragment => {
    ...
    this.route.queryParams.subscribe(queryParams => {
        ...
    });
});
```

Then, we make sure the display type is set before we parse the query params.

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.html
jmedinaebi: I managed to fix this issue.

1. You may have some CSS changes locally, revert them as what you have here is not far from working.

2. Update classes in table-interactor-column.component.css
    ```
    .interactorsColumn {
      position: sticky;
      left: 0;
      z-index: 3;
      max-height: 45px;
      min-height: 45px;
      min-width: 130px;
      max-width: 130px;
      padding: 5px;
      text-align: end;
      color: #0e6f76;
      font-weight: bold;
      background: inherit;
    }
    
    .interactorsColumnFixedWidth {
      width: 130px;
    }
    
    .intStoich {
      height: 45px;
      max-width: 70px;
      min-width: 70px;
      font-size: medium;
      text-align: end;
      z-index: 1;
    }

    .intStoichFixedWidth {
      width: 70px;
    }
    ```

3. Update html to use new CSS classes
    3.1. lines 15-16
    ```
    <td class="interactorsColumn" [ngClass]="complexes.length<=8 ? 'horizontalHeader' : 'interactorsColumnFixedWidth'"
            [style]="ranges.length!=0 ? 'left:3ch':''">
    ```
    3.2. line 39
    ````
    <td class="intStoich" [ngClass]="{'intStoichFixedWidth': complexes.length > 8}">
    ````
    3.3. line 82
    ```
    <td class="intStoich" [ngClass]="{'intStoichFixedWidth': complexes.length > 8}">
    ```
    3.4. also line 91, not classes but there is a bug there
    ```
    <div class="spaceHolder" *ngIf="complexes.length>8"></div>
    ```

jmedinaebi: I tested it with the searches
- nuclear pore
- KMN: 9 results
- CPX-4441: 3 results

src/app/complex/complex-results/complex-results.component.ts
```diff
@@ -194,4 +242,47 @@ export class ComplexResultsComponent implements OnInit, AfterViewInit {
...
+    if (this._complexSearch.elements.length === 1) {
+      const complexId = this._complexSearch.elements[0].complexAC;
+      if (!!complexId) {
+        // For some reason this is needed so the navigate call works
+        this.router.routeReuseStrategy.shouldReuseRoute = function () {
+          return false;
+        };
+        this.router.navigate(['/complex', complexId]);
+      }
+    } else if (this._complexSearch.elements.length < this._navigatorPageSize) {
```
jmedinaebi: A few things here.

1. `this._complexSearch.elements` are the elements in the page, so they will always be 15 or 20. We should do 
```
if (this._complexSearch.totalNumberOfResults <= this._navigatorPageSize) {
```

2. As you can see in the code above, we should use `<=` instead of `<`. If there are 20 results or less, we want the navigator view.

3. More importantly, the way this is set up, we always check the length of the elements after a search. That means that if someone goes to `http://localhost:4200/complex/search?query=KMN&page=1#view_list`, which has the fragment `#view_list`, because there are 9 results, they get redirected to the navigator view. We should only do this if the fragment is not set, which happens for urls like `http://localhost:4200/complex/search?query=KMN&page=1` or when people initially search. To fix this we need to move this if/else inside the if below:
```
    if (!this.DisplayType) {
      if (this._complexSearch.totalNumberOfResults <= this._navigatorPageSize) {
        this.setComplexNavigatorView();
      } else {
        this.setListView();
      }
    }
```

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-name/table-interactor-name.component.html
```diff
@@ -0,0 +1,36 @@
...
+    <div class="name" title="More information about {{interactorId}}" *ngIf="!IDDisplay">
+      <a *ngIf="!IDDisplay" href="{{identifierLink}}" target="_blank"
+         class="externalLinkContainer">
+        {{ interactorName }}
+      </a>
+    </div>
+
+    <a [routerLink]="['/complex/search']"
+       [queryParams]="{query: interactorId, page: 1}"
+       [preserveFragment]="true"
```
jmedinaebi: Because we now check the number of results to decide between list or navigator view, it makes sense to get rid of this line. So, instead preserving the fragment, when people click on search of one of the interactors, it opens a new tab, and that tab is list or navigator view depending on the number of results.

src/app/complex/complex-results/complex-results.component.ts
```diff
@@ -5,39 +5,56 @@ import {ComplexPortalService} from '../shared/service/complex-portal.service';
...
-
+    this._allInteractorsInComplexSearch = [];
+    this.route.fragment.subscribe(fragment => {
+      if (fragment === this.COMPLEX_NAVIGATOR_VIEW) {
+        this.DisplayType = this.COMPLEX_NAVIGATOR_VIEW;
+      } else if (fragment === this.LIST_VIEW) {
+        this.DisplayType = this.LIST_VIEW;
+      }
+    });
     this.route
```
jmedinaebi: Here I didn't mean to move `this.route.queryParams.subscribe` after `this.route.fragment.subscribe`. I meant moving it inside. The potential issue with the 2 subscribers separate is that they are synchronous, so we cannot make sure of the order they run, so we may be checking the query params before the fragment and we shouldn't as there are variables, like page index, that depend on the fragment. So, to avoid that we can do

```
    this.route.fragment.subscribe(fragment => {
      if (fragment === this.COMPLEX_NAVIGATOR_VIEW) {
        this.DisplayType = this.COMPLEX_NAVIGATOR_VIEW;
      } else if (fragment === this.LIST_VIEW) {
        this.DisplayType = this.LIST_VIEW;
      }
      this.route
        .queryParams
        .subscribe(queryParams => {
            ...
        });
    });
```

So, after processing the fragment, we process the query params.

src/app/complex/complex-results/complex-navigator/complex-navigator-rating/complex-navigator-rating.component.html
```diff
@@ -0,0 +1,15 @@
+<div class="ratingComponent" *ngIf="visible===true">
+  <div class="text">
+    <p>Tell us about your experience with the Complex Navigator!</p>
+    <button class="close" id="closePopup" (click)="ratingVisibility()" style="color: white">&times;</button>
```
jmedinaebi: The 'x' to close the pop up is not positioned right, it should be in the top right corner, and it's next to the text.

Screenshot 2024-05-23 at 14 01 45

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/complex-navigator-utils.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export function formatStoichiometryValues(stoichiometry: string): string {
+  const parsedStoichiometry = parseStoichiometryValues(stoichiometry);
+  if (!!parsedStoichiometry) {
+    const minValue = parsedStoichiometry[0];
+    const maxValue = parsedStoichiometry[1];
+    if (minValue === maxValue) {
+      return minValue.toString();
+    } else {
+      return `${minValue}, ${maxValue}`;
```
jmedinaebi:
```
return `${minValue}-${maxValue}`;
```

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-main-interactor/table-main-interactor.component.css
```diff
@@ -0,0 +1,17 @@
...
+  margin-right: auto;
+  min-width: 1%;
+  max-width: 1%;
+  max-height: 80%;
+  min-height: 80%;
+  z-index: 0;
+}
+
+.verticalLine {
+  border-right: 5px solid #0e6f76;
```
jmedinaebi: Change 5px to 3px to make the lines thiner.
```
  border-right: 3px solid #0e6f76;
```

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-subcomponent-interactor/table-subcomponent-interactor.component.css
```diff
@@ -0,0 +1,17 @@
...
+  margin-left: auto;
+  margin-right: auto;
+  min-width: 1%;
+  max-width: 1%;
+  max-height: 50%;
+  min-height: 50%;
+  z-index: 0;
+}
+
+.verticalLine {
```
jmedinaebi, line 11: Same as above, change 5px to 3px to make the lines thiner.
```
  border-right: 3px solid #0e6f76;
```

src/app/complex/complex-results/complex-results.component.ts
```diff
@@ -194,4 +242,47 @@ export class ComplexResultsComponent implements OnInit, AfterViewInit {
...
+    this.DisplayType = this.COMPLEX_NAVIGATOR_VIEW;
+    this.reloadPage();
+    if (!!this.toast) {
+      this.notificationService.closeAnnouncement(this.toast.toastId);
+      this.toast = null;
+    }
+  }
+
+
+  setFirstDisplayType(): void {
```
jmedinaebi, line 266: To fix the redirect, what we need to do is to only redirect when people do a search. When people do a search, we don't have the display type set up, then we can redirect. When people change a filter, or change pages, or change from list to navigator or the other way around, display type is set, and we don't need to redirect.

So, that means, moving the redirect logic to inside the if `if (!this.DisplayType) {`

Basically, this method ends up like this
```
  setFirstDisplayType(): void {
    if (!this.DisplayType) {
      if (this._complexSearch.totalNumberOfResults === 1) {
        const complexId = this._complexSearch.elements[0].complexAC;
        if (!!complexId) {
          // For some reason this is needed so the navigate call works
          this.router.routeReuseStrategy.shouldReuseRoute = function () {
            return false;
          };
          this.router.navigate(['/complex', complexId]);
        }
      } else {
        // Currently the list view is the default, as we are just launching the navigator view
        // Later on we can change the default view to be the list or navigator view based on number of results
        if (this._complexSearch.totalNumberOfResults <= this._navigatorPageSize) {
          this.setComplexNavigatorView();
        } else {
          this.setListView();
        }
      }
    }
  }
```

src/app/complex/complex-results/complex-results.component.ts
```diff
@@ -5,44 +5,61 @@ import {ComplexPortalService} from '../shared/service/complex-portal.service';
...
-  private _pageSize = 10;
   private _spicesFilter: string[];
   private _bioRoleFilter: string[];
   private _interactorTypeFilter: string[];
+  private _allInteractorsInComplexSearch: Interactor[] = [];
+  DisplayType: string;
+  toast;
+
+  private _listPageSize = 15; // This is where we set the size of the pages for list view
+  private _navigatorPageSize = 15; // This is where we set the size of the pages for navigator view
```
jmedinaebi: Do we want to leave the paginator size in 15? Or something larger?

src/app/complex/complex-results/complex-results.component.ts
```diff
@@ -153,22 +192,31 @@ export class ComplexResultsComponent implements OnInit, AfterViewInit {
 
   set complexSearch(value: ComplexSearchResult) {
     this._complexSearch = value;
+    this.setFirstDisplayType(); // if only one complex is in the search result, the display is list
```
jmedinaebi: This comment is outdated. If only one complex, then we redirect to the detailed complex page

PR #217, opened 2024-06-06
Fix expanded shadow bug

jmedinaebi:
When there is only one component in the subcomplex, ngClass was setting the class as both `firstExpandedRow` and `lastExpandedRow`, but they were not working together.

Instead I created a new CSS class for the case when we have one subcomponent and we want the show before and after that row. To simplify the ngClass logic, I moved the logic to a method in the ts file and simply call that method from the ngClass in the HTML.

Unrelated to this, but I saw it while testing this, if the organism name was null or undefined, `formatOrganismName` was throwing an error.

### DIFF
--- a/src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.css
+++ b/src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.css
@@ -100,6 +100,15 @@ a, a:visited {
   background: linear-gradient(0deg, #0e6f7696 0px, transparent 10px);
 }
 
+.singleExpandedRow:after {
+  z-index: 40;
+  content: '';
+  position: absolute;
+  inset: -2px -1px -2px -1px;
+  pointer-events: none;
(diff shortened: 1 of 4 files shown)


src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.css
```diff
@@ -100,6 +100,15 @@ a, a:visited {
   background: linear-gradient(0deg, #0e6f7696 0px, transparent 10px);
 }
 
+.singleExpandedRow:after {
+  z-index: 40;
+  content: '';
+  position: absolute;
+  inset: -2px -1px -2px -1px;
+  pointer-events: none;
+  background: linear-gradient(180deg, #0e6f7696 0px, transparent 10px), linear-gradient(0deg, #0e6f7696 0px, transparent 10px);
```
jmedinaebi, line 109: New class has both linear gradients, the one for before the row, and the one for after.

PR #225, opened 2024-08-14
Upgrade angular

EliotRagueneau:
Update of Complex Portal including:

- [X] Angular 12 -> 18
- [X] RxJs 6.6.7 -> 7.4.0
- [X] Migration to new application builder based on esbuild https://v17.angular.io/guide/esbuild
- [X] Migration to signals for inputs and outputs https://angular.dev/guide/signals
  - [ ] Would ideally need to update ComplexNavigator to optimise it using the signals


EliotRagueneau:
The migrations are kind of optionals, I'm fully open to debate over the interest of them with you, in order to establish our standards

EliotRagueneau:
The application builder is actually quite a lot faster to compile. It also creates a really small improvement in the loading time of the home page, but it's still not great tbh

EliotRagueneau:
The usage of signals for `input` and `output` is genuinly great I think to simplify the reactivity of components to input changes, and also grant us some nice little improvements like the `model`. However, while I did an automatic migration from `@Input` to `input<T>()`, it's not taking advantage of the signal strength as for that, we would need to create the useful `computed()` values, but I think having the base signal is already great, and then we can progressively take advantage of the strengths of signals when we develop further the different components.

The ComplexNavigator would benefit a lot from this I think, but the code is quite complicated to update (I did give a try, but it's big mess tbh)

BTW that made me see that @susiehgt you named some variables not according to standards, with capital case for the first later. Be careful about that please ... It was particularly present around the display options

src/app/basket/basket.component.html
```diff
@@ -11,7 +11,7 @@ <h2>Personal collection of favourite complexes
     <h4>{{ getKeys(complexBasket).length }} complexes in your basket</h4>
     <div class="row">
       <cp-complex-list-display-buttons
-        [displayType]="DisplayType"
```
EliotRagueneau, line 14: @susiehgt bad naming

src/app/basket/basket.component.ts
```diff
@@ -21,7 +22,7 @@ export class BasketComponent implements OnInit, AfterViewInit {
   private _complexBasket: { [name: string]: BasketItem };
   complexSearchBasket: ComplexSearchResult = null;
   allInteractorsInComplexSearchBasket: Interactor[] = [];
-  DisplayType: string;
```
EliotRagueneau, line 24: @susiehgt bad naming

src/app/basket/basket.component.ts
```diff
@@ -34,26 +35,24 @@ export class BasketComponent implements OnInit, AfterViewInit {
   ngOnInit() {
     this.titleService.setTitle('Complex Portal - Basket');
     this.complexNavigatorLoading();
-    this.route.fragment.subscribe(fragment => {
+    this.route.fragment.pipe(take(1)).subscribe(fragment => {
```
EliotRagueneau, line 37: Avoid memory leaks by releasing the subscription after the initial change

src/app/basket/basket.component.ts
```diff
@@ -34,26 +35,24 @@ export class BasketComponent implements OnInit, AfterViewInit {
...
       if (fragment === COMPLEX_NAVIGATOR_VIEW) {
-        this.DisplayType = COMPLEX_NAVIGATOR_VIEW;
```
EliotRagueneau, line 39: @susiehgt bad naming

src/app/basket/basket.component.ts
```diff
@@ -34,26 +35,24 @@ export class BasketComponent implements OnInit, AfterViewInit {
...
+        this.displayType = COMPLEX_NAVIGATOR_VIEW;
       } else {
-        this.DisplayType = LIST_VIEW;
```
EliotRagueneau, line 41: @susiehgt bad naming

src/app/complex/complex-results/complex-navigator-buttons/complex-navigator-buttons.component.ts
```diff
@@ -1,57 +1,52 @@
...
-  @Output() IDDisplayChange = new EventEmitter<boolean>();
+  interactorsSorting = model<string>();
+  organismIconDisplay = model<boolean>();
+  interactorTypeDisplay = model<boolean>();
+  idDisplay = model<boolean>();
 
-  @Input() interactorsSorting: string;
-  @Input() organismIconDisplay: boolean;
-  @Input() interactorTypeDisplay: boolean;
-  @Input() IDDisplay: boolean;
```
EliotRagueneau, line 17: @susiehgt bad naming

src/app/basket/basket.module.ts
```diff
@@ -3,15 +3,15 @@ import {CommonModule} from '@angular/common';
...
+import {ProgressSpinnerComponent} from '../shared/loading-indicators/progress-spinner/progress-spinner.component';
 
 @NgModule({
   imports: [RouterModule.forChild([
     {path: '', component: BasketComponent}
   ]),
     CommonModule,
     ComplexNavigatorModule,
-    ProgressSpinnerModule
+    ProgressSpinnerComponent
```
jmedinaebi, line 14: Why the component and not the module? We are importing `CommonModule` and `ComplexNavigatorModule`. Why import `ProgressSpinnerComponent` instead of `ProgressSpinnerModule`?

src/app/complex/complex-details/complex-details.module.ts
```diff
@@ -38,22 +38,22 @@ import {ComplexPortalService} from '../shared/service/complex-portal.service';
...
+import {ProgressSpinnerComponent} from '../../shared/loading-indicators/progress-spinner/progress-spinner.component';
 
 @NgModule({
   imports: [
     RouterModule.forChild([
       {path: '', component: ComplexDetailsComponent},
     ]),
     CommonModule,
-    ProgressSpinnerModule,
+    ProgressSpinnerComponent,
```
jmedinaebi, line 56: Same as above.

src/app/complex/complex-details/complex-header/download-modal/download-modal.component.ts
```diff
@@ -1,50 +1,34 @@
...
 
   constructor(private googleAnalyticsService: AnalyticsService) {
   }
 
   ngOnInit() {
   }
-
-
-  goToComplexWSJsonFormat(): void {
-    this.googleAnalyticsService.fireDownloadResourceEvent(Category.details, 'ComplexWS');
```
jmedinaebi, line 22: Not that it's particularly relevant, but we're losing the Google Analytics events here. I prefer using `href` in the HTML instead of window.open, but if we care about the Google analytics, we're missing that.

Also, we're missing the `_blank` property now.

src/app/complex/complex-organisms/complex-organisms.module.ts
```diff
@@ -3,15 +3,15 @@ import {CommonModule} from '@angular/common';
...
 import {MatButtonModule} from '@angular/material/button';
+import {ProgressSpinnerComponent} from '../../shared/loading-indicators/progress-spinner/progress-spinner.component';
 
 @NgModule({
   imports: [RouterModule.forChild([
     {path: '', component: ComplexOrganismsComponent}
   ]),
     CommonModule,
-    ProgressSpinnerModule,
+    ProgressSpinnerComponent,
```
jmedinaebi, line 14: Same as above.

src/app/complex/complex-results/complex-navigator/complex-list-display-buttons/complex-list-display-buttons.component.html
```diff
@@ -1,11 +1,11 @@
...
-  <button type="submit" name="submit" class="button view-button"
-          (click)="setListView()" [ngClass]="{'selected': !isDisplayComplexNavigatorView()}">
+  <button class="button view-button"
+          (click)="setListView()" [class.selected]="displayType() === 'view_list'">
     View list
   </button>
-  <button type="submit" name="submit" class="button view-button cnButton"
-          (click)="setComplexNavigatorView()" [ngClass]="{'selected': isDisplayComplexNavigatorView()}">
+  <button class="button view-button cnButton"
+          (click)="setComplexNavigatorView()" [class.selected]="displayType() === 'view_complex_navigator'">
```
jmedinaebi: The point of using the function `isDisplayComplexNavigatorView` is to only define the string in a single place and then rely in the const. Now the `view_complex_navigator` is in two places in the HTML and two places in the ts file, so there's more changes to have a typo, and if we update it, more changes to miss one of the places to change it.

src/app/complex/complex-results/complex-navigator/complex-navigator.module.ts
```diff
@@ -21,16 +21,16 @@ import {
...
 import {ComplexListDisplayButtonsComponent} from './complex-list-display-buttons/complex-list-display-buttons.component';
 import {MatTooltipModule} from '@angular/material/tooltip';
+import {ProgressSpinnerComponent} from '../../../shared/loading-indicators/progress-spinner/progress-spinner.component';
 
 @NgModule({
   imports: [
     RouterModule,
     CommonModule,
-    ProgressSpinnerModule,
+    ProgressSpinnerComponent,
```
jmedinaebi, line 33: Same as above.

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
```diff
@@ -34,21 +34,22 @@ export class EnrichedComplex {
...
-  @Input() IDDisplay: boolean;
-
+  complexes = input<Element[]>([]);
+  interactorsSorting = input<string>();
+  interactors = input<Interactor[]>([]);
+  organismIconDisplay = input<boolean>(true);
+  interactorTypeDisplay = input<boolean>(true);
+  idDisplay = input<boolean>(true);
+
+  // TODO rework bellow to compute when needed from the inputs
```
jmedinaebi, line 44: This already happens, right? We only call `enrichInteractors` in ngOnChanges if there are changes in interactors input.

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-name/table-interactor-name.component.ts
```diff
@@ -9,22 +8,22 @@ import {organismIcon} from '../../../../../complex-portal-utils';
...
-  @Input() IDDisplay: boolean;
-  @Input() isMainInteractor: boolean;
+  interactorName = input<string>();
+  interactorType = input<string>();
+  interactorId = input<string>();
+  interactorOrganism = input<string>();
+  identifierLink = input<string>();
+  organismIconDisplay = input<boolean>();
+  interactorTypeDisplay = input<boolean>();
+  IDDisplay = input<boolean>();
```
jmedinaebi: idDisplay?

src/app/complex/complex-results/complex-results.component.ts
```diff
@@ -41,44 +48,45 @@ export class ComplexResultsComponent implements OnInit, AfterViewInit {
...
+    this.route.fragment.pipe(take(1)).subscribe(fragment => {
       if (fragment === COMPLEX_NAVIGATOR_VIEW) {
-        this.DisplayType = COMPLEX_NAVIGATOR_VIEW;
-      } else if (fragment === LIST_VIEW) {
-        this.DisplayType = LIST_VIEW;
+        this.displayType = COMPLEX_NAVIGATOR_VIEW;
+      } else {
+        this.displayType = LIST_VIEW;
       }
-      this.route
```
jmedinaebi, line 51: There was a reason `this.route.queryParams.subscribe` was inside `this.route.fragment.subscribe`. 

`this.requestComplexResults();` calls the back-end with a param for the page size. The page size depends on the view mode (list or navigator), which may depend on the fragment set.

By having the subscribe one inside the other, we make sure if there is a fragment, it is set properly before making the back-end call. If fragment is null, `this.route.fragment.subscribe` still works, I tested that.

Now there are 2 separate subscribe calls that can happen in any order.

Also, we had
```
      if (fragment === COMPLEX_NAVIGATOR_VIEW) {
        this.DisplayType = COMPLEX_NAVIGATOR_VIEW;
      } else if (fragment === LIST_VIEW) {
        this.DisplayType = LIST_VIEW;
      }
```
instead of 
```
      if (fragment === COMPLEX_NAVIGATOR_VIEW) {
        this.displayType = COMPLEX_NAVIGATOR_VIEW;
      } else {
        this.displayType = LIST_VIEW;
      }
```
because list view is not the default view. If the fragment is specified, then we use that, if not, when we process the search results, based on the number of results, we set one or the other.

src/app/complex/complex-results/complex-results.module.ts
```diff
@@ -5,20 +5,20 @@ import {RouterModule} from '@angular/router';
...
+import {ProgressSpinnerComponent} from '../../shared/loading-indicators/progress-spinner/progress-spinner.component';
 
 
 @NgModule({
   imports: [RouterModule.forChild([
     {path: '', component: ComplexResultsComponent},
   ]),
     CommonModule,
-    ProgressSpinnerModule, MarkdownModule,
+    ProgressSpinnerComponent, MarkdownModule,
```
jmedinaebi: Same as above.

src/app/basket/basket.module.ts
jmedinaebi: I see you deleted the module. Shouldn't it then go in declarations instead of imports?

EliotRagueneau: I made the ProgressSpinnerComponent a standalone component as that was essentially what it already was, just made it official let's say

src/app/complex/complex-details/complex-header/download-modal/download-modal.component.ts
EliotRagueneau: I initially wanted to make the link visible and accessible to users, but actually that makes the download not work since https://stackoverflow.com/questions/49474775/chrome-65-blocks-cross-origin-a-download-client-side-workaround-to-force-down, it creates more issue since it just opens it in a new tab if its coming from a cross origin. So I'm reverting back to a click action instead. Thanks for pointing that out!

EliotRagueneau: fixed in [eef8b69](https://github.com/Complex-Portal/complex-portal-view/pull/225/commits/eef8b69584a6e77db2a8ba2706a1a3862b7319cf)

src/app/complex/complex-results/complex-navigator/complex-list-display-buttons/complex-list-display-buttons.component.html
EliotRagueneau: Fair, though I introduced a type restriction which is stronger than a recommended const usage. we can probably mix both by using an enum instead

EliotRagueneau: Fixed in [e9e9501](https://github.com/Complex-Portal/complex-portal-view/pull/225/commits/e9e9501c495d5febf02f421555eae1753eac0ac1)

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
EliotRagueneau: Yes in principle indeed, but ngOnChanges is called for any changes, not specifically for the concerned inputs.
Using signals would be quite powerful here, but as stated in my comments, not so straightforward  as the code is kinda convoluted over this component.

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-name/table-interactor-name.component.ts
EliotRagueneau: Forgot about this one, fixing it

EliotRagueneau: Fixed in [6396889](https://github.com/Complex-Portal/complex-portal-view/pull/225/commits/639688955c4b61e90af191972bd83866b8d9044c)

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
jmedinaebi: That's why there's an if inside ngOnChanges. ngOnChanges is called for any changes, but it has an argument with the changes. So we check if the field `interactors` is part of the changes to call `enrichInteractors`.

There may cleaner ways to do it with the Angular upgrade, but we're not currently calling `enrichInteractors` every time `ngOnChanges` is called.

src/app/complex/complex-results/complex-results.component.ts
EliotRagueneau: Basically the problem was that if the fragment is null, then displayType is null which made that no button was selected, even though list view was being rendered. In general I think it's better practice to avoid null values in this case, but I'm fine for this case if you think it's really better. 

Regarding the nested subsribes, I don't think that's a good thing as you will create many subsrctiptions which will never be unsubsrcribed. I'm switching it to a fragment -> tap -> switchMap(queryParam) -> subsribe. This way only one subscribe, and basically you retrigger the requestCompelxResults() every time the fragment change, after having set the right view with the tap.

However, regarding the setting of the view according to the number of results, I couldn't find this logic, and I don't think it exists, otherwise, we wouldn't have null value for displayType.

EliotRagueneau: See [2cf650f](https://github.com/Complex-Portal/complex-portal-view/pull/225/commits/2cf650f5f97f110705c9369e2d845f4fa21edae6) to consider if the changes suits you or not

jmedinaebi: It happens in `processSearchResults`.

If we have no filters and there's 1 result, then we redirect to the single complex page.
Else, if displayType is not set (that's why it can be null initially), if we have less or equal to 20 results, then displayType is set as navigator, otherwise as list.

If the fragment is specified, then displayType is set from the fragment, if it's not, it will be based on the number of results.

jmedinaebi: Somehow it's not working and it's never setting the default view now. It's always list view by default.

jmedinaebi: I think I know why. `displayType` is used here
```
<cp-complex-list-display-buttons
              [displayType]="displayType"
              (displayTypeChange)="onDisplayTypeChange($event)">
            </cp-complex-list-display-buttons>
```
Then, in that component 
```
displayType = model<SearchDisplay>(SearchDisplay.list);
```

So by default is list, and then the method above does not work, as displayType is never null or undefined.

jmedinaebi: I've fixed it locally. I actually broke it some months ago...

In `complex-list-display-buttons.component.ts`:
```
displayType = model<SearchDisplay>();
```

In `complex-results.component.ts`:
- Delete line 218
   ```
   this.displayType = displayType;
   ```
- Add to the beginning of `setListView()`
   ```
   this.displayType = SearchDisplay.list;
   ```
- Add to the beginning of `setComplexNavigatorView()`
   ```
   this.displayType = SearchDisplay.navigator;
   ```

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
EliotRagueneau: Yes yes of course, just saying it would be a really nice place to use the new signals, but indeed it is not required

src/app/complex/complex-results/complex-results.component.ts
EliotRagueneau: Done in [c6baad2](https://github.com/Complex-Portal/complex-portal-view/pull/225/commits/c6baad2446f15a17111cada2dcb6c9d33e9992d3)

src/app/complex/complex-details/complex-header/complex-header.component.css
```diff
@@ -5,5 +5,18 @@
 
 h2 {
   overflow-wrap: break-word;
+  display: -webkit-box;
```
jmedinaebi, line 8: Should we use the truncate pipe here for the name? Same as we do for the name in the list view? Or does the CSS properties work better here?

src/app/complex/complex-results/complex-list/complex-list.component.css
```diff
@@ -21,3 +22,12 @@
 .basket-icons {
   margin-left: 5px;
 }
+
+.predicted h3 {
+  font-family: "Roboto Mono", monospace;
```
jmedinaebi, line 27: I'm not sure about the different fonts.

Screenshot 2024-08-16 at 08 35 21

jmedinaebi:
I just pushed a couple of commit with some small changes I had on a different branch. Now we got everything in this one.

src/app/complex/complex-details/complex-header/complex-header.component.css
EliotRagueneau: The css version is definitively better as it doesn't require a mono font to specify a number of line, but it yielded to long dashed lines which were extending after the end of text, which was pretty ugly

src/app/complex/complex-results/complex-list/complex-list.component.css
EliotRagueneau: It's definitely arguable as a choice, basically I needed a mono font to have consistent char width to get precisely 2 lines max, and then Roboto is pretty close to what we have, but different enough to mark the difference (+ I found it funny to use Roboto for machine predictions ^^).

Of course we could style the font at least to have the same size and weight as the curated ones, but I do think it convey the idea that these are somehow weaker. 

I'm perfectly willing to remove this though, or at least to tweak it so that it blends better, if the majority of people don't like it :D

src/app/complex/complex-details/complex-header/download-modal/download-modal.component.html
```diff
@@ -7,7 +7,7 @@ <h3>Download</h3>
   <div class="row button-grid small-up-1 medium-up-4 large-up-4 margin-top-xlarge no-pad-right"
        data-equalize-on="medium" id="large-button-grid">
     <div class="column  padding-bottom-large" *ngFor="let format of formats()">
-      <a class="button medium-12 columns text-center" [class.disabled]="format.disabled" (click)="download(format.url, format.filename)">
+      <a class="button medium-12 columns text-center" [class.disabled]="format.disabled" (click)="download(format)">
```
EliotRagueneau, line 10: Good change indeed !

src/assets/styles/css/complex-portal-theme.scss
```diff
@@ -295,3 +276,30 @@ a:visited {
 }
 
 /*********** END HOME PAGE SECTION *******************************/
+
+
+.predicted {
```
EliotRagueneau, line 281: I grouped every predicted special case in a single parent using scss, it was more handy than adding `.predicted` before every item ^^

EliotRagueneau:
I think we should merge this PR before everything happens here. I think we're all good for the transition to angular 18 right @jmedinaebi ? Did you face any challenge with the installation of new deps or anything ?

jmedinaebi:
> I think we should merge this PR before everything happens here. I think we're all good for the transition to angular 18 right @jmedinaebi ? Did you face any challenge with the installation of new deps or anything ?

Yes, let's merge it before it gets even bigger with more changes. I had no issue with the installation, I had to update my node version, but that was it.

PR #226, opened 2024-08-21
Move logic from components to classes

jmedinaebi:
Move logic from the following components to new classes for `EnrichedComplex` and `EnrichedInteractors`:
- `table-interactor-column.component.ts`
- `table-main-interactor.component.ts`
- `table-subcomponent-interactor.component.ts`

I moved the methods to:
- calculate the time each interactor appears
- sort interactors
- toggle interactors
- find interactor (or subcomponent) in complex
- calculate start and end indexes
- check if lines start, cross or end at a interactor for a complex

I created a class to have all enriched interactors, to encapsulate all the methods to calculate ranges, sort, etc. Now `table-interactor-column.component.ts` is so much cleaner and small.

When we add orthology groups, we can do something similar. Have a new class for the orthology group, and the methods for that, like a method to check if a complex contains a protein in a given orthology group, can be in that class, instead of having all these methods in the components. We would also need to add new methods in EnrichedComplex to check when the lines start and end for the orthology groups.

I also got rid of `startInteractorIncludedWhenExpanded`, as we don't really need it.


EliotRagueneau:
In principle, why not, I'm just slightly alarmed by the spearation of concerns: how is it that it is the Complex which should know where the it's line start and stops? This only makes sense in the context of a table, but by itself, it doesn't have anything to do with it right ?

EliotRagueneau:
It's true components are much cleaner though !

PR #229, opened 2024-09-04
Sort complexes to always display curated first

jmedinaebi:
From the NAR paper (https://docs.google.com/document/d/1hZj-HhGyn6BuWfK9_9GaEQA7Xl15nTpDFq2VNwhqyBI/edit) Sucharita is writing:
```
Search results in the Navigator View display manually curated complexes first,
and ordered such that complexes with the most components are listed first.
```
It is described we display curated complexes before predicted complexes in the navigator view.

And Henning also commented on that, saying it is not the current behaviour, but it should be:
```
not currently, but hopefully this will be implemented?
```

I have made some changes to make it behave like that and this is how it looks for a search like `Q8TAD8 Q96A72 Q9H307 P38919 Q6NUS6` (https://complex-portal.github.io/complex-portal-view/complex/search?query=Q8TAD8%20Q96A72%20Q9H307%20P38919%20Q6NUS6&page=1). 

#### Current order (in develop branch, deployed in the github pages)
Screenshot 2024-09-04 at 12 13 01

#### New order
Screenshot 2024-09-04 at 12 13 14

### DIFF
--- a/src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
+++ b/src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
@@ -88,7 +88,8 @@ export class TableInteractorColumnComponent implements OnChanges {
         subComponents: null,
         partOfComplex: [],
         timesAppearing: 0,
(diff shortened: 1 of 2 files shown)


EliotRagueneau:
This PR motivated me to improve the clustering, so I created #230 which is solving the same issue, but using similarity score instead of creating new variables / system

PR #230, opened 2024-09-04
Improve clustering

EliotRagueneau:
In reaction to #229 , I reimplemented the clustering of elements by following https://stackoverflow.com/a/64338609

Performances are a bit impacted, but still acceptable (roughly 200ms to sort complexes).
Improvements are particularly visible when several well identified clusters are visible, like with `Q06217 Q14974 Q16514 Q16552 Q16594 Q24572 Q4VC05 Q60973` on 
https://complex-portal.github.io/complex-portal-view/complex/search?query=Q06217%20Q14974%20Q16514%20Q16552%20Q16594%20Q24572%20Q4VC05%20Q60973#view_navigator
compared to http://localhost:4200/complex/search?query=Q8TAD8%20Q96A72%20Q9H307%20P38919%20Q6NUS6#view_navigator

The order of elements are now firstly sorted by which complex they appear in first, so that clusters are more compacted. Then I use the timeAppearing, but in the other direction than before: the least commons elements are first, then come the most common ones. That is an arguable decision, but it allows us to limit long lines. Could be something to do UX testing on to see what people prefer, because this makes cluster appear lower on the page.

### DIFF
--- a/src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
+++ b/src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
@@ -17,6 +17,7 @@ export class EnrichedInteractor {
   subComponents: ComplexComponent[];
   partOfComplex: number[];
   timesAppearing: number;
+  indexAppearing: number;
 }
 
 export class EnrichedComplex {
@@ -85,6 +86,7 @@ export class TableInteractorColumnComponent implements OnChanges {
         subComponents: null,
         partOfComplex: [],
         timesAppearing: 0,
+        indexAppearing: this.complexes().findIndex(complex => complex.componentAcs.has(interactor.identifier))
       };
       if (isSubComplex) {
         this.loadSubInteractors(newEnrichedInteractor).subscribe(subComponents => newEnrichedInteractor.subComponents = subComponents);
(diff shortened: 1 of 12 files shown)


EliotRagueneau:
I did not push to github page yet so that you can see the before / after easily

EliotRagueneau:
Also I used tensorflow for the matrix manipulations, but it might be overkill and overweigth for our needs. We could try to implement it using another lighter lib. I chose it initially because it is GPU accelerated, but unsure it's really worth the lib weight

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
```diff
@@ -234,74 +236,27 @@ export class TableInteractorColumnComponent implements OnChanges {
...
+  private compareFn = (a: EnrichedInteractor, b: EnrichedInteractor) =>
+    -(b.indexAppearing - a.indexAppearing) || // First by order of appearance (staircase effect)
+    -(b.timesAppearing - a.timesAppearing); // Then by reversed occurrence, in order to minimize edge length
 
-  public classifyInteractorsByType() {
-    this.enrichedInteractors.sort((a, b) => {
-      if (b.interactor.interactorType === a.interactor.interactorType) {
-        return b.timesAppearing - a.timesAppearing;
-      } else {
-        const typeBTimesAppearing = this._timesAppearingByType.get(b.interactor.interactorType);
```
jmedinaebi, line 259: This logic using `_timesAppearingByType` and `_timesAppearingByOrganism` was to display first the types or organism with the most complexes. Now we're back to sorting them by name.

The changes in sort by occurrence makes sense, also the change to sort interactors with the same type or organism; but this other change we should not do.

jmedinaebi: If I search for "*", and then group by type; currently the proteins go first and then the small molecules (at the bottom of the table). https://complex-portal.github.io/complex-portal-view/complex/search?query=*#view_navigator

Screenshot 2024-09-05 at 08 41 04

With these changes the small molecules are displayed first (at the top of the table) and then the proteins, as we simply sort the types alphabetically.

Screenshot 2024-09-05 at 08 41 15

EliotRagueneau:
Nicely spotted, this new commit fixes this issue, while refactoring a bit to avoid redundancies

EliotRagueneau:
I think it's all good now, okay for merge @jmedinaebi ?

PR #232, opened 2024-09-18
Use interface and classes to support ortholog groups

jmedinaebi:
A few things in this PR:
- Renaming interactor as component everywhere
- Remove references to subcomplexes where possible and just mention subcomponents
- Use enums for sorting and grouping types
- Rename `EnrichedComplex` as `NavigatorComplex`
- Rename `EnrichedInteractor` as `INavigatorComponent`
- Create 2 classes that implement `INavigatorComponent`, one for simple components and one for ortholog groups
- Move logic to create `INavigatorComponent` to before we do the clustering, so we do the clustering after the ortholog groups are created


src/app/complex/complex-results/complex-navigator-buttons/complex-navigator-buttons.component.ts
```diff
@@ -1,24 +1,41 @@
...
 export class ComplexNavigatorButtonsComponent implements OnInit {
-  interactorsSorting = model<string>();
+  componentsSorting = model<NavigatorComponentSorting>();
+  componentsGrouping = model<NavigatorComponentGrouping>();
   organismIconDisplay = model<boolean>();
   interactorTypeDisplay = model<boolean>();
   idDisplay = model<boolean>();
+  complexSearch = input<ComplexSearchResult>();
+
+  orthologButtonAvailable = computed(() => this.checkIfPantherGroups());
```
jmedinaebi: Try this instead.

1. Delete method `checkIfPantherGroups`

2. Update `orthologButtonAvailable` to be
   ```
   orthologButtonAvailable = input<boolean>();
   ```

3. In `complex-navigator.component.ts` create a new boolean variable to store if there are ortholog groups
   ```
   orthologGroupsAvailable = true;
   ```

4. And pass it to the complex-navigator-buttons component in `complex-navigator-component.html`
   ```
    <cp-complex-navigator-buttons
      [(componentsSorting)]="componentsSorting"
      [(componentsGrouping)]="componentsGrouping"
      [(organismIconDisplay)]="organismIconDisplay"
      [(interactorTypeDisplay)]="interactorTypeDisplay"
      [(idDisplay)]="idDisplay"
      [orthologButtonAvailable]="orthologGroupsAvailable"
      [complexSearch]="complexSearch()">
    </cp-complex-navigator-buttons>
   ```

5. In `table-structure.component.ts` create a model to set the ortholog groups value
   ```
   orthologGroupsAvailable = model<boolean>();
   ```

6. And in that same component, set the value in the `createOrthologGroups` method if we create any group
   ```
    ...
    let anyOrthologGroupCreated = false; // new line
    const newNavigatorComponents: INavigatorComponent[] = [];
    for (const [_, proteins] of Object.entries(groupedByOrthology)) {
      if (proteins.length > 1) {
        // Multiple proteins in the ortholog group
        const group = proteins[0].orthologsGroup;
        const orthologsGroup = new NavigatorOrthologGroup(group, proteins);
        newNavigatorComponents.push(orthologsGroup);
        anyOrthologGroupCreated = true; // new line
      } else if (proteins.length === 1) {
        // Single proteins in the ortholog group, there's no need to create the group, we just use the protein component
        newNavigatorComponents.push(proteins[0]);
      }
    }
    this.orthologGroupsAvailable.set(anyOrthologGroupCreated); // new line
    newNavigatorComponents.push(...interactorsWithoutGroup);
    return newNavigatorComponents;
   ```

7. Finally, in `complex-navigator.component.html` pass the `orthologGroupsAvailable` variable to the model so it can be set
   ```
   <cp-table-structure [complexSearch]="complexSearch()"
                      [interactors]="interactors()"
                      [componentsSorting]="componentsSorting"
                      [componentsGrouping]="componentsGrouping"
                      [organismIconDisplay]="organismIconDisplay"
                      [interactorTypeDisplay]="interactorTypeDisplay"
                      [idDisplay]="idDisplay"
                      [canAddComplexesToBasket]="canAddComplexesToBasket()"
                      [canRemoveComplexesFromBasket]="canRemoveComplexesFromBasket()"
                      [(orthologGroupsAvailable)]="orthologGroupsAvailable"
                      (onComplexRemovedFromBasket)="onComplexRemovedFromBasket.emit($event)">
   </cp-table-structure>
   ```

I haven't tried this but it should work. `TableStructureComponent` sets if there are ortholog groups or not, it passes the value through the model to `ComplexNavigatorComponent` which passes it to `ComplexNavigatorButtonsComponent` to enable or disable the button.

It's like what we do for the sorting or grouping, but the other way around.

package.json
```diff
@@ -37,6 +37,7 @@
     "expose-loader": "^5.0.0",
     "jquery": "^3.7.1",
     "litemol": "https://github.com/dsehnal/LiteMol/tarball/master",
+    "marked": "^12.0.2",
```
jmedinaebi: What's this dependency for?

src/app/complex/complex-results/complex-navigator/table-structure/model/navigator-ortholog-group.model.ts
```diff
@@ -0,0 +1,127 @@
...
+    this._subComponents = subComponents;
+    this._hidden = false;
+    this._expanded = false;
+    this._timesAppearing = 0;
+    this._indexAppearing = 0;
+    this._isOrthologGroup = true;
+  }
+
+  get isOrthologGroup(): boolean {
+    return this._isOrthologGroup;
```
jmedinaebi: We can just do `return true;`

src/app/complex/complex-results/complex-navigator/table-structure/model/navigator-simple-component.model.ts
```diff
@@ -0,0 +1,118 @@
...
+    this._hidden = false;
+    this._expanded = false;
+    this._timesAppearing = 0;
+    this._indexAppearing = 0;
+    this._subComponents = null;
+    this._isOrthologGroup = false;
+  }
+
+  get isOrthologGroup(): boolean {
+    return this._isOrthologGroup;
```
jmedinaebi: We can just do `return false;`

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.html
```diff
@@ -84,28 +85,28 @@
...
+                    [identifierLink]="subComponent.identifierLink"
                     [organismIconDisplay]="organismIconDisplay()"
                     [interactorTypeDisplay]="interactorTypeDisplay()"
                     [idDisplay]="idDisplay()"
                   ></cp-table-interactor-name>
                 </div>
               </td>
-              @for (complex of enrichedComplexes; track complex.complex.complexAC) {
+              @for (complex of navigatorComplexes; track complex.complex.complexAC) {
                 <td class="intStoich" [ngClass]="{'intStoichFixedWidth': complexes().length > 6}">
```
jmedinaebi: You need to update this td element to support predicted subcomponents
```
                <td class="intStoich" [ngClass]="{'intStoichFixedWidth': complexes().length > 6}"
                    [class.predicted]="complex.complex.predictedComplex">
```

src/app/complex/complex-results/complex-navigator-buttons/complex-navigator-buttons.component.ts
```diff
@@ -1,24 +1,40 @@
...
+import {ComplexSearchResult} from '../../shared/model/complex-results/complex-search.model';
 
 @Component({
   selector: 'cp-complex-navigator-buttons',
   templateUrl: './complex-navigator-buttons.component.html',
   styleUrls: ['./complex-navigator-buttons.component.css']
 })
 export class ComplexNavigatorButtonsComponent implements OnInit {
-  interactorsSorting = model<string>();
+  componentsSorting = model<NavigatorComponentSorting>();
```
jmedinaebi: We don't need this anymore, as we have the output. You can delete this, delete `this.componentsSorting.set(typeOfSorting);` on line 37 and delete `[(componentsSorting)]="componentsSorting"` in complex-navigator.component.html (line 4).

jmedinaebi:
In the basket, update the complex navigator component.

In basket.component.html (line 52)
```
              <cp-complex-navigator
                [complexSearch]="complexSearchBasket"
                [interactors]="allInteractorsInComplexSearchBasket"
                [canAddComplexesToBasket]="false"
                [canRemoveComplexesFromBasket]="true"
                [(componentsGrouping)]="componentsGrouping"
                [(componentsSorting)]="componentsSorting"
                (onComplexRemovedFromBasket)="deleteComplexFromBasket($event)">
              </cp-complex-navigator>
```

And then in basket.component.ts, add
```
  componentsGrouping = NavigatorComponentGrouping.DEFAULT;
  componentsSorting = NavigatorComponentSorting.DEFAULT;

```

package.json
susiehgt: My code would not compile if I did not add it, the console was asking me to

jmedinaebi:
Regarding adding the type of display to the parameters, we should also add a parameter for view vs navigator view, and remove the fragment.

To do that, in `complex-results-component.ts` we do:

1. Update `ngOnInit` to not parse from the fragment, but to get the display mode from the params
   ```
   ngOnInit() {
    this.titleService.setTitle('Complex Portal - Results');
    this.allInteractorsInComplexSearch = [];
    this.route.queryParams.pipe().subscribe(queryParams => {
      this.query = queryParams['query'];
      Object.keys(this.filters).forEach(filter => this.filters[filter] = this.decodeURL(filter, queryParams));
      // Only set the mode if in the params, otherwise rely on the default behaviour
      if (!!queryParams['displayMode']) {
        this._displayType = queryParams['mode'];
      }
      this.confidenceFilter = queryParams['minConfidence'] ? Number(queryParams['minConfidence']) : 1;
      this.componentsGrouping = queryParams['componentsGrouping'] ? queryParams['componentsGrouping'] : NavigatorComponentGrouping.DEFAULT;
      this.componentsSorting = queryParams['componentsSorting'] ? queryParams['componentsSorting'] : NavigatorComponentSorting.DEFAULT;
      this.typeOfDisplay = queryParams['typeOfDisplay'] ? queryParams['typeOfDisplay'] : NavigatorDisplayType.DETAILED;
      this.currentPageIndex = queryParams['page'] ? Number(queryParams['page']) : 1;
      // TODO This is out for now, but CP-84 (JIRA )should fix that!!
      // this.pageSize = queryParams['size'] ? Number(queryParams['size']) : 10;
      this.requestComplexResults();
      document.body.scrollTop = 0;
    });
   }
   ```

2. Update `reloadPage` to set a new param for the type of display
   ```
   queryParams['displayMode'] = this._displayType;
   ```

3. Update `reloadPage` to remove setting the fragment on the navigate call
   ```
   this.router.navigate([], { queryParams });
   ```

jmedinaebi:
And to add the display options to the params, do:

1. Create a new enum
   ```
   export enum NavigatorDisplayType {
     COMPACT = 'compact',
     DETAILED = 'detailed'
   }
   ```

2. Set the new type in `complex-results.component.ts`, same as we do for the grouping and sorting. A new variable, setting it from the params in ngOnInit, setting the param in reloadPage and a new method like `onComplexNavigatorGroupingChange` for the new variable.

3. Update `ComplexNavigatorComponent` with a new model for the new variable, same as `componentsSorting` or `componentsGrouping` and a new method like `onDisplayTypeChange`.

4. Update `ComplexNavigatorButtonsComponent` with a new input `typeOfDisplay = input<NavigatorDisplayType>();`

5. Remove `ngOnInit` from `ComplexNavigatorButtonsComponent`

6. Add a new method, like `onGroupingChanged` to `ComplexNavigatorButtonsComponent` for the new variable.

7. Update where `typeOfDisplay` is used in `ComplexNavigatorButtonsComponent` as it is now an input and requires (), and also where it is set, as we should simply call the new on??? method.

8. Replace strings 'compact' and 'detailed' with the enum.

src/app/complex/complex-results/complex-navigator-buttons/complex-navigator-buttons.component.html
```diff
@@ -19,14 +30,14 @@
       <li>
         <label>
           <input type="radio" name="displayType" value="Compact" class="checkbox compact"
-                 (click)="setDisplayType('compact')" [checked]="typeOfDisplay === 'compact'">
+                 (click)="setDisplayType('compact')" [checked]="typeOfDisplay() === NavigatorDisplayType.COMPACT">
```
jmedinaebi: Replace
```
setDisplayType('compact')
```

With 
```
setDisplayType(NavigatorDisplayType.COMPACT)
```

Same for line 40.

jmedinaebi: We also need to update the type of the argument in `setDisplayType`
```
setDisplayType(type: NavigatorDisplayType) {
```

And update the ifs in that method to use the enum.

src/app/complex/complex-results/complex-navigator/complex-navigator.component.ts
```diff
@@ -15,8 +24,139 @@ export class ComplexNavigatorComponent {
   canRemoveComplexesFromBasket = input<boolean>();
   onComplexRemovedFromBasket = output<string>();
 
-  interactorsSorting = 'Occurrence';
+  componentsSorting = model<NavigatorComponentSorting>();
+  componentsGrouping = model<NavigatorComponentGrouping>();
+  displayType = model<NavigatorDisplayType>();
```
jmedinaebi: These 3 can be input instead of model. We already have outputs to emit events, so we don't need to use model here.

jmedinaebi: We would need to update the html (basket and complex-results) where `cp-complex-navigator` element is used to use `[` instead of `[(`.

src/app/complex/complex-results/complex-navigator/complex-navigator.component.ts
```diff
@@ -15,8 +24,139 @@ export class ComplexNavigatorComponent {
...
+  onGroupingChange = output<NavigatorComponentGrouping>();
+  onSortingChange = output<NavigatorComponentSorting>();
+  onDisplayTypeChange = output<NavigatorDisplayType>();
+
   organismIconDisplay = true;
   interactorTypeDisplay = true;
   idDisplay = true;
+
+  _allChanges: OutputRef<any>[] = [
```
jmedinaebi: No longer used, delete.

src/app/complex/complex-results/complex-navigator/complex-navigator.component.ts
```diff
@@ -15,8 +24,139 @@ export class ComplexNavigatorComponent {
...
+  constructor(private complexPortalService: ComplexPortalService) {
+    effect(() => {
+      this.setNavigatorComponents(this.navigatorComponentsGroupedByOrthologs(), this.navigatorComponentsWithoutGrouping());
+      this.setIconsDisplay();
+    });
+    this._allChanges.forEach(change => change.subscribe(() => setTimeout(() => this.anyChange.emit())));
+  }
+
+  onGroupingChanged(componentsGrouping: NavigatorComponentGrouping) {
+    this.componentsGrouping.set(componentsGrouping);
```
jmedinaebi: Remove this set call. Same for the set calls in `onSortingChanged` and `onDisplayTypeChanged`. We send the events and then the other components update this, so we don't need it.

src/app/complex/complex-results/complex-results.component.ts
```diff
@@ -47,19 +50,14 @@ export class ComplexResultsComponent implements OnInit, AfterViewInit {
...
-      switchMap(_ => this.route.queryParams)
-    ).subscribe(queryParams => {
+    this.route.queryParams.pipe().subscribe(queryParams => {
       this.query = queryParams['query'];
       Object.keys(this.filters).forEach(filter => this.filters[filter] = this.decodeURL(filter, queryParams));
       this.confidenceFilter = queryParams['minConfidence'] ? Number(queryParams['minConfidence']) : 1;
+      this.componentsGrouping = queryParams['componentsGrouping'] ? queryParams['componentsGrouping'] : NavigatorComponentGrouping.DEFAULT;
+      this.componentsSorting = queryParams['componentsSorting'] ? queryParams['componentsSorting'] : NavigatorComponentSorting.DEFAULT;
+      this.typeOfDisplay = queryParams['typeOfDisplay'] ? queryParams['typeOfDisplay'] : NavigatorDisplayType.DETAILED;
+      this._displayType = queryParams['displayMode'] ? queryParams['displayMode'] : queryParams['mode'];
```
jmedinaebi:
```
      if (!!queryParams['displayMode']) {
        this._displayType = queryParams['displayMode'];
      }
```

src/app/complex/complex-results/complex-navigator/complex-navigator.component.ts
jmedinaebi: Actually ignore this, leave them as model.

src/app/complex/complex-results/complex-navigator/complex-navigator.component.ts
jmedinaebi: Ignore this, we can simplify all this logic. As mentioned above, keep the models. Then, remove the 3 outputs `onGroupingChange`, `onSortingChange` and `onDisplayTypeChange`; and have only 1 output `anyChange = output<void>();`

src/app/complex/complex-results/complex-navigator/complex-navigator.component.ts
jmedinaebi: Ignore, do not remove the set calls. Keep them, but replace the `this.onGroupingChange.emit(componentsGrouping);` call with `this.anyChange.emit();`. Same on the 2 other methods below.

src/app/complex/complex-results/complex-results.component.html
```diff
@@ -43,19 +43,25 @@ <h2 class="padding-left-large">Total number of results: {{ complexSearch.totalNu
...
+                                      [complexSearch]="complexSearch"
+                                      [interactors]="allInteractorsInComplexSearch"
+                                      [canAddComplexesToBasket]="true"
+                                      [canRemoveComplexesFromBasket]="false"
+                                      [(componentsGrouping)]="componentsGrouping"
+                                      [(componentsSorting)]="componentsSorting"
+                                      [(displayType)]="typeOfDisplay"
+                                      (onGroupingChange)=onComplexNavigatorGroupingChange(componentsGrouping)
+                                      (onSortingChange)=onComplexNavigatorSortingChange(componentsSorting)
+                                      (onDisplayTypeChange)="onComplexNavigatorDisplayChange(typeOfDisplay)">
```
jmedinaebi: Replace these 3 lines with 
```
(anyChange)=reloadPage()>
```

`reloadPage` needs to be made public.

src/app/complex/complex-results/complex-results.component.ts
```diff
@@ -136,6 +137,21 @@ export class ComplexResultsComponent implements OnInit, AfterViewInit {
     this.reloadPage();
   }
 
+  public onComplexNavigatorGroupingChange(grouping: NavigatorComponentGrouping) {
```
jmedinaebi: Delete these 3 methods. No longer needed after the other changes mentioned on other comments.

src/app/complex/complex-results/complex-navigator/table-structure/model/navigator-ortholog-group.model.ts
```diff
@@ -0,0 +1,127 @@
...
+import {XRef} from '../../../../shared/model/complex-results/interactor.model';
+
+export class NavigatorOrthologGroup implements INavigatorComponent {
+  private _orthologXref: XRef;
+  private _hidden: boolean;
+  private _expanded: boolean;
+  private _subComponents: INavigatorComponent[];
+  private _timesAppearing: number;
+  private _indexAppearing: number;
+  private _isOrthologGroup: boolean;
```
jmedinaebi: We already return always true, we don't need this variable.

src/app/complex/complex-results/complex-navigator/table-structure/model/navigator-simple-component.model.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export class NavigatorSimpleComponent implements INavigatorComponent {
+  private _interactor: Interactor;
+  private _hidden: boolean;
+  private _isSubComplex: boolean;
+  private _expanded: boolean;
+  private _subComponents: ComplexComponent[];
+  private _timesAppearing: number;
+  private _indexAppearing: number;
+  private _isOrthologGroup: boolean;
```
jmedinaebi: We already return always false, we don't need this variable.

src/app/complex/complex-results/complex-navigator/complex-navigator.component.ts
```diff
@@ -13,10 +22,132 @@ export class ComplexNavigatorComponent {
   interactors = input<Interactor[]>();
   canAddComplexesToBasket = input<boolean>();
   canRemoveComplexesFromBasket = input<boolean>();
+  componentsSorting = model<NavigatorComponentSorting>();
+  componentsGrouping = model<NavigatorComponentGrouping>();
+  displayType = model<NavigatorDisplayType>();
   onComplexRemovedFromBasket = output<string>();
+  onGroupingChange = output<NavigatorComponentGrouping>();
+  onSortingChange = output<NavigatorComponentSorting>();
+  onDisplayTypeChange = output<NavigatorDisplayType>();
```
jmedinaebi: These 3 outputs can be deleted, no longer used.

jmedinaebi (commented):
I managed to get the search for panther xrefs working. It is not one of the default fields we search for, but it works if we specify the field. Meaning, if we search for `PTHR11973` we get nothing, but if we search for `pxref:("PTHR11973")` we do.

So, to make this work nicely.

1. Update `INavigatorSubComponent` with a new get method.
   ```
   get componentQuery(): string;
   ```

2. Update `NavigatorSimpleComponent` to just return the id on that method.
   ```
   get componentQuery(): string {
     return this.identifier;
   }

   ```

3. Update `NavigatorOrthologGroup` to return the proper query we need
   ```
   get componentQuery(): string {
     return 'pxref:("' + this.identifier + '")';
   }
   ```

4. Update `table-interactor-name.component` component with a new input
   ```
   interactorQuery = input<string>();
   ```

5. Update the html of that component to use the input on the search (lines 41 and 60)
   ```
   [queryParams]="{query: interactorQuery(), page: 1}"
   ```

6. Update `table-interactor-column.component` component to pass the new field to the component updated before
   6.1. For the main interactors (line 26)
     ```
     ...
              [interactorId]="interactor.identifier"
              [interactorQuery]="interactor.componentQuery"
     ...
     ```
   6.2. For the subcomponents we can use the identifier as we currently do
     ```
     ...
                    [interactorId]="subComponent.identifier"
                    [interactorQuery]="subComponent.identifier"
     ...
     ```

src/app/complex/complex-results/complex-navigator/table-structure/model/navigator-simple-component.model.ts
```diff
@@ -0,0 +1,120 @@
...
+  set indexAppearing(value: number) {
+    this._indexAppearing = value;
+  }
+
+  componentIds(): string[] {
+    return [this.identifier];
+  }
+
+  get orthologsGroup(): XRef {
+    for (const xref of this._interactor.xrefs) {
```
jmedinaebi: In case we want to deploy these changes before the back-end changes, which we might, we need to update this.

The back-end (production one) is not returning xrefs now, so `interactor.xrefs` would be undefined. So we need to do:

```
    if (!!this._interactor.xrefs) {
      for (const xref of this._interactor.xrefs) {
        if (!!xref && xref.database === 'panther') {
          return xref;
        }
      }
    }
    return null;
```

Basically adding the if at the beginning to be safe.

jmedinaebi (commented):
Really good, just one small thing before we're ready to merge this.

src/app/app.component.ts
```diff
@@ -85,7 +85,7 @@ export class AppComponent implements OnInit, AfterViewInit {
   //       localMasthead.className += ' meta-background-color';
   //     }
   //     if (localMastheadImage != null) {
-  //       localMasthead.style.backgroundImage = 'url(' + localMastheadImage.getAttribute("content") + ')';
+  //       localMasthead.style.backgroundImage = 'state(' + localMastheadImage.getAttribute("content") + ')';
```
jmedinaebi: I'm guessing you did a replace all and this got replace by accident. It is not relevant as it is commented out, but revert it anyway.

src/app/basket/basket.component.css
```diff
@@ -1,9 +1,3 @@
-.ComplexNavigator, .hover {
```
jmedinaebi, line 1: Why was this removed?

src/app/complex/complex-results/complex-filter/complex-filter.component.ts
```diff
@@ -43,7 +43,10 @@ export class ComplexFilterComponent implements OnInit {
 
   onResetAllFilters = output<boolean>();
 
-  confidenceGap = input<number>(2);
+  /**
```
jmedinaebi, line 46: Can you add a bit more? Minimum 5, otherwise it breaks? What's 5?

src/app/complex/complex-results/complex-navigator/table-structure/model/navigator-component.model.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  set hidden(value: boolean);
+
+  set expanded(value: boolean);
+
+  set timesAppearing(value: number);
+
+  set indexAppearing(value: number);
+
+  componentIds(): string[];
```
jmedinaebi: To be consistent, make it `get componentIds(): string[];`. Then you need to replace where it is called, from `componentIds()` to `componentIds`. And also change the classes that implement this interface.

src/app/complex/complex-results/complex-navigator/service/state/complex-navigator-display.service.ts
```diff
@@ -0,0 +1,66 @@
...
+    typeIconDisplay: this.typeIconDisplay(),
+  }));
+
+  ignore = false;
+
+  constructor(private route: ActivatedRoute, private router: Router) {
+    this.route.queryParams.subscribe(queryParams => {
+      for (const param of Object.keys(this.params())) {
+        if (!queryParams[param]) {
+          continue;
```
jmedinaebi: We can read from local storage here. If there's no query params for a param, then we fetch from local storage.

jmedinaebi: The other part is to save to the local storage. That I'm not sure how at the moment.

src/app/complex/complex-results/complex-results.component.ts
```diff
@@ -32,34 +31,26 @@ export class ComplexResultsComponent implements OnInit, AfterViewInit {
...
   private _toast;
-  private _listPageSize = 15; // This is where we set the size of the pages for list view
-  private _navigatorPageSize = 20; // This is where we set the size of the pages for navigator view
-  private _listCurrentPage: number;
-  private _navigatorCurrentPage: number;
-  private _listLastPageIndex;
-  private _navigatorLastPageIndex;
-
-  constructor(private route: ActivatedRoute, private router: Router,
+  private pageSize = 20; // This is where we set the size of the pages for navigator view
```
jmedinaebi: Comment is no longer true. We're now using page size 20 for both views.

src/app/complex/complex-results/complex-navigator/table-structure/model/navigator-simple-component.model.ts
```diff
@@ -0,0 +1,116 @@
...
+  set indexAppearing(value: number) {
+    this._indexAppearing = value;
+  }
+
+  componentIds(): string[] {
+    return [this.identifier];
+  }
+
+  get orthologsGroup(): XRef {
+    return this._interactor.xrefs?.find(xref => xref.qualifier === 'orthology-group');
```
jmedinaebi, line 110: Just qualifier? No check for the database?

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.html
```diff
@@ -46,66 +38,60 @@
...
                     [i]="i"
                     [j]="j"
-                    [enrichedInteractors]="enrichedInteractors"
+                    [navigatorComponents]="navigatorComponents()"
                   ></cp-table-subcomponent-interactor>
                 </td>
               }
-              @if (enrichedComplexes.length > 6) {
+              @if (navigatorComplexes.length > 6) {
                 <div class="spaceHolder"></div>
```
jmedinaebi, line 86:
```
<div [class.sorting]="isSorting()" class="spaceHolder"></div>
```

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.html
```diff
@@ -22,22 +22,14 @@
           <td class="interactorsColumn"
```
jmedinaebi:
```
<td class="interactorsColumn horizontalHeader"
              [class.fixedWidth]="fixedWidth()"
              [style]="ranges.length!=0 ? 'left:3ch':''">
```

jmedinaebi: The CSS file also needs to be updated.

Currently, it is like this, which it is not working.
```
.horizontalHeader {
  min-width: 114px;
  max-width: 114px;
}

.horizontalHeader.fixedWidth {
  width: 230px;
}
```

It works if you do:
```
.horizontalHeader {
  min-width: 114px;
  max-width: 114px;
}

.horizontalHeader.fixedWidth {
  width: 230px;
  min-width: 230px;
  max-width: 230px;
}
```

jmedinaebi (commented):
Looking good. I left a few comments on some small things.

src/app/basket/basket.component.css
susiehgt: Eliot used the medium-12 in the HTML so I think he removed it because it is not used anymore

src/app/complex/complex-results/complex-filter/complex-filter.component.ts
susiehgt: @EliotRagueneau

src/app/basket/basket.component.css
EliotRagueneau: Yes basically before, we were limiting the size of the lements themeselves, but I neeeded to restrict the size of the prent really, so that the tabs match the content. Then i could have just moved that up, but basically, 70% can be problematic when on small screens, so I used indeed teh small-12 medium-10 large-8 I think to ensure that in all scales, we have enough space

src/app/complex/complex-results/complex-filter/complex-filter.component.ts
EliotRagueneau: Yep, basically the slider has a minimum length that I couldn't overwrite, at least easily, so the minimum space between stars in order for the star pattern to take as much space as the slider is 5. If you go bellow, the area gets ddimisnished, but not the slider, leading it to be over the next area, and not alligend with the stars.
5 is pixel size

jmedinaebi: Then the comment should be something like
```
Slider length cannot be overwritten. In order for the stars to take the same space as the slider, the minimum space between stars must be 5, otherwise they don't align.
```

src/app/complex/complex-results/complex-navigator/service/state/complex-navigator-display.service.ts
EliotRagueneau: We can just add it to the other effect we have to write to the localStorage

src/app/complex/complex-results/complex-results.component.ts
EliotRagueneau: Yes we actually removed that because it was a lot of complexity and reloading time just to have 5 iteems less in the list view, which is not that important in my opinion

jmedinaebi: That makes sense, but the comment is not updated.

src/app/complex/complex-results/complex-navigator/table-structure/model/navigator-simple-component.model.ts
EliotRagueneau: I prefer to use the qualifier if we decide to change the source orthology database at some point, since they are many alternatives, and there is no "clear winner" of who's the best

src/app/complex/complex-results/complex-navigator/service/state/complex-navigator-display.service.ts
```diff
@@ -0,0 +1,95 @@
...
+          }
+          // 2. Then heck local storage
+        } else if (localStorage.getItem(param) !== null) {
+          switch (typeof this[param]()) {
+            case 'boolean':
+              value = localStorage.getItem(param) === 'true';
+              break;
+            default:
+              value = localStorage.getItem(param);
+          }
```
jmedinaebi: We can create a private method parseParam
```
private parseParam(paramName, paramValue): any {
  switch (typeof this[paramName]()) {
    case 'boolean':
      return paramValue === 'true';
    default:
      return paramValue;
  }
}
```

And then, simplify this code:
```
        // 1. Check URL params first
        if (queryParams[param] !== undefined) {
          value = this.parseParam(param, queryParams[param]);
          // 2. Then heck local storage
        } else if (localStorage.getItem(param) !== null) {
          value = this.parseParam(param, localStorage.getItem(param));
        } else {
          ...
```

src/app/complex/complex-results/complex-navigator/service/state/complex-navigator-display.service.ts
```diff
@@ -0,0 +1,95 @@
...
+          value = this[param]();
+        }
+        if (value !== this[param]()) {
+          this[param].set(value);
+        }
+      }
+    });
+
+    effect(() => {
+      if (this.ignore) {
```
jmedinaebi, line 62: What's the ignore variable?

src/app/complex/complex-results/complex-navigator/service/state/complex-navigator-display.service.ts
```diff
@@ -0,0 +1,95 @@
...
+        return;
+      }
+      this.router.navigate([], {queryParams: this.params(), queryParamsHandling: 'merge'});
+    }, {allowSignalWrites: true});
+
+
+    effect(() => {
+      this.updateLocalStorage();
```
jmedinaebi: Instead of having 2 effects, we could just have one, right?

```
    effect(() => {
      if (this.ignore) {
        return;
      }
      this.updateLocalStorage();
      this.router.navigate([], {queryParams: this.params(), queryParamsHandling: 'merge'});
    }, {allowSignalWrites: true});
```

src/app/shared/google-analytics/service/analytics.service.ts
```diff
@@ -70,4 +70,12 @@ export class AnalyticsService {
   public rateComplexNavigator(rating: number) {
     this.analytics.event(Action[Action.RatingComplexNavigator], Category[Category.complexNavigator], 'rating', rating);
   }
+
+  public useOrthologyButton(action: Action) {
```
jmedinaebi: Rename this method, as it is more generic than ortholog button. It can be `useComplexNavigatorSettingButton`.

jmedinaebi: Rename this method, as it is more generic than ortholog button. It can be useComplexNavigatorSettingButton.

src/app/complex/complex-results/complex-navigator/service/state/complex-navigator-display.service.ts
```diff
@@ -0,0 +1,95 @@
...
+        // 1. Check URL params first
+        if (queryParams[param] !== undefined) {
+          switch (typeof this[param]()) {
+            case 'boolean':
+              value = queryParams[param] === 'true';
+              break;
+            default:
+              value = queryParams[param];
+          }
+          // 2. Then heck local storage
```
jmedinaebi: Typo, I guess you meant `Then check...`

jmedinaebi:
Apart from the comments above, a few other things regarding the table style.

1. Something changed about the width that broke the table. I think it happens with and without orthologs, and with or without sorting.
Screenshot 2024-10-01 at 10 53 59

2. The shadow of the header extends all the way to the right
Screenshot 2024-10-01 at 10 54 06

3. The shadow is not displayed when we're at the very top of the table, so the separation between the header and the first row is not easy to see
Screenshot 2024-10-01 at 10 54 17

I'll take a look to see if I can find out how to fix these issues, so you can focus on the xml maker.

EliotRagueneau:
Screenshot 2024-10-01 at 12 58 26
I think the fix you did on shadows made things even worse

jmedinaebi:
> Screenshot 2024-10-01 at 12 58 26 I think the fix you did on shadows made things even worse

I see what's happening. I think I know why it's happening, so I'll take a look.

EliotRagueneau:
Normally you don't have to precise the shadow's width, it's just folllowing its container size, so all you neeed to do is have the parent the right size really

jmedinaebi:
> Normally you don't have to precise the shadow's width, it's just folllowing its container size, so all you neeed to do is have the parent the right size really

That's what I was thinking on doing.

EliotRagueneau:
@jmedinaebi thanks for the fix on groups, completely missed that! 
I did change quite a lot of the calculations in the css to only do it in the parrent and use the result everywhere. I'm not sure it's actually optimising since it looks like the variable contains the calc and not the result, but at least I find it clearer and avoid code duplication :D
I think we can aceept this PR now, would you agree?

jmedinaebi:
> @jmedinaebi thanks for the fix on groups, completely missed that! I did change quite a lot of the calculations in the css to only do it in the parrent and use the result everywhere. I'm not sure it's actually optimising since it looks like the variable contains the calc and not the result, but at least I find it clearer and avoid code duplication :D I think we can aceept this PR now, would you agree?

I think it is ready.

@EliotRagueneau @susiehgt 
Complex indexer has finished and I have updated SOLR in the public instances too. Public Complex web services are now returning orthology groups, and I have deployed these changes on the github pages with the production environment instead of the test one. You can now check these changes with production data, like this search https://complex-portal.github.io/complex-portal-view/complex/search?query=*&page=1&minConfidence=1&displayMode=navigator-view&mergeOrthologs=true&componentsSorting=Default&idDisplay=true&organismIconDisplay=true&typeIconDisplay=true&predicted=true

PR #233, opened 2024-10-02
:sparkles: Set measures of rotation based on angle, width and height of the tilted headers

EliotRagueneau:
Make the rotation based on the angle, the width of a column, and the visible height of a column header. 

This enables the filling of space available by the navigator, but it has not been implemented yet.

I've made the choice to perform the calculation in js instead of using css variables as their values are being evaluated at usage, and not at declaration, meaing that we would be calculating many many time trigonometric functions if we are using css variables. Unfortunately, using js makes it less easy to play with params, and will make the calculation of the desired column width slightly harder.

On the bonus side, I've also fixed the spaceholder at the end of the line so that all columns behave the same on hover

### DIFF
--- a/src/app/basket/basket.component.html
+++ b/src/app/basket/basket.component.html
@@ -10,7 +10,7 @@
       </h2>
       <h4>{{ getKeys(complexBasket).length }} complexes in your basket</h4>
       <div class="row expanded">
-        <mat-tab-group [tabIndex]="displayType === SearchDisplay.list ? 0 : 1"
+        <mat-tab-group [tabIndex]="displayType === SearchDisplay.list ? 0 : 1" #tabs
                        (selectedTabChange)="onDisplayTypeChange($event.index === 0 ? SearchDisplay.list : SearchDisplay.navigator)"
                        class="columns medium-centered small-12 medium-10 large-8">
           <mat-tab label="List view">
--- a/src/app/basket/basket.component.html
+++ b/src/app/basket/basket.component.html
@@ -43,7 +43,6 @@
             </table>
           </mat-tab>
           <mat-tab label="Navigator view">
-            <div class="columns medium-12">
               @if (complexSearchBasket) {
                 <div class="ComplexNavigator"
                      [ngClass]="complexSearchBasket.totalNumberOfResults <=6 ? 'smallCN' : ''">
@@ -56,7 +55,6 @@
               } @else {
                 <cp-progress-spinner></cp-progress-spinner>
               }
-            </div>
           </mat-tab>
         </mat-tab-group>
       </div>
--- a/src/app/basket/basket.component.html
+++ b/src/app/basket/basket.component.html
@@ -12,7 +12,7 @@
       <div class="row expanded">
         <mat-tab-group [tabIndex]="displayType === SearchDisplay.list ? 0 : 1"
                        (selectedTabChange)="onDisplayTypeChange($event.index === 0 ? SearchDisplay.list : SearchDisplay.navigator)"
-                       class="columns medium-centered small-12 medium-10 large-8">
+                       class="columns medium-centered small-12 medium-10 large-10">
           <mat-tab label="List view">
             <table class="hover">
               <thead>
--- a/src/app/basket/basket.component.html
+++ b/src/app/basket/basket.component.html
@@ -10,7 +10,7 @@
       </h2>
       <h4>{{ getKeys(complexBasket).length }} complexes in your basket</h4>
       <div class="row expanded">
-        <mat-tab-group [tabIndex]="displayType === SearchDisplay.list ? 0 : 1" #tabs
+        <mat-tab-group [tabIndex]="displayType === SearchDisplay.list ? 0 : 1"
                        (selectedTabChange)="onDisplayTypeChange($event.index === 0 ? SearchDisplay.list : SearchDisplay.navigator)"
                        class="columns medium-centered small-12 medium-10 large-10">
           <mat-tab label="List view">
--- a/src/app/complex/complex-results/complex-navigator/complex-navigator.component.ts
+++ b/src/app/complex/complex-results/complex-navigator/complex-navigator.component.ts
@@ -28,16 +28,12 @@ export class ComplexNavigatorComponent implements AfterViewInit {
   navigatorComponentsWithoutGrouping = computed(() => this.createNavigatorComplexes(this.complexSearch().elements, this.interactors()));
   navigatorComponentsGroupedByOrthologs = computed(() => this.createOrthologGroups(this.navigatorComponentsWithoutGrouping()));
   orthologGroupsAvailable = computed(() => this.navigatorComponentsGroupedByOrthologs().some(c => c instanceof NavigatorOrthologGroup));
-  isSorted = computed(() => this.state.componentsSorting() !== NavigatorComponentSorting.DEFAULT);
-
-  sortingWidth = computed(() => this.isSorted() ? 20 : 0);
-  interactorWidth = computed(() => 230 + this.sortingWidth());
 
   @ViewChild('parent') parent: ElementRef<HTMLDivElement>;
   navigatorComponents: INavigatorComponent[] = [];
 
   constructor(private complexPortalService: ComplexPortalService, public state: NavigatorStateService) {
-    effect(() => this.numberOfColumns() && this.adjustColWidth(), {allowSignalWrites: true});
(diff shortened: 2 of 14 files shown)


EliotRagueneau:
To understand better the names and math behind the formulas, I have created this little geogebra which illustrate the math behind the formulas, in case we need to tweak them at some point: https://www.geogebra.org/calculator/jtx3rcs4

src/app/complex/complex-results/complex-navigator/complex-navigator.component.ts
```diff
@@ -30,9 +30,8 @@ export class ComplexNavigatorComponent {
   navigatorComponents: INavigatorComponent[] = [];
 
   constructor(private complexPortalService: ComplexPortalService, public state: NavigatorStateService) {
-    effect(() => {
-      this.setNavigatorComponents(this.navigatorComponentsGroupedByOrthologs(), this.navigatorComponentsWithoutGrouping());
-    });
+    effect(() => this.state.numberOfColumns.set(this.complexSearch().elements.length), {allowSignalWrites: true});
```
EliotRagueneau: Why not using computed instead of an effect for this?

jmedinaebi: How can we use computed? `numberOfColumns` is in the `NavigatorStateService` so we need to call `this.state.numberOfColumns.set`. I don't know how to use computed to do that.

EliotRagueneau:
I've made the same calculation as you, but on the measured size of the component, and I update this measure when the window is resized, when the number of columns evolve and when the tab is selected (otherwise total is 0 if the tab was previously disselected)

jmedinaebi:
> I've made the same calculation as you, but on the measured size of the component, and I update this measure when the window is resized, when the number of columns evolve and when the tab is selected (otherwise total is 0 if the tab was previously disselected)

I was having a look at your latest commit. I will take a better look, but it looks way more complicated than before. What's the actual visible or performance change? Before we had 1 effect, now we still that effect, but we have many other things. Is the difference now that we don't have the total width size? We calculate based on the window?

src/app/basket/basket.component.html
```diff
@@ -10,9 +10,9 @@ <h2>Personal collection of favourite complexes
       </h2>
       <h4>{{ getKeys(complexBasket).length }} complexes in your basket</h4>
       <div class="row expanded">
-        <mat-tab-group [tabIndex]="displayType === SearchDisplay.list ? 0 : 1"
+        <mat-tab-group [tabIndex]="displayType === SearchDisplay.list ? 0 : 1" #tabs
```
jmedinaebi: What's `#tabs`?

src/app/complex/complex-results/complex-results.component.html
```diff
@@ -33,14 +33,14 @@ <h2 class="padding-left-large">Total number of results: {{ complexSearch.totalNu
...
+                             (selectedTabChange)="onTabChange($event, list, navigatorTab)">
                 <mat-tab label="List view" #list>
                   <cp-complex-list [complexSearch]="complexSearch"/>
                 </mat-tab>
-                <mat-tab label="Navigator view" #navigator>
+                <mat-tab label="Navigator view" #navigatorTab>
                   <div class="ComplexNavigator"
                        [ngClass]="complexSearch.totalNumberOfResults <=6 ? 'smallCN' : 'largeCN'">
-                    <cp-complex-navigator class="Complex-navigator"
+                    <cp-complex-navigator class="Complex-navigator" #navigator
```
jmedinaebi: What's `#navigator`?

src/app/complex/complex-results/complex-results.component.html
```diff
@@ -33,14 +33,14 @@ <h2 class="padding-left-large">Total number of results: {{ complexSearch.totalNu
               </cp-complex-paginator>
 
               <mat-tab-group [selectedIndex]="isDisplayComplexNavigatorView() ? 1 : 0"
-                             (selectedTabChange)="onTabChange($event, list, navigator)">
+                             (selectedTabChange)="onTabChange($event, list, navigatorTab)">
                 <mat-tab label="List view" #list>
```
jmedinaebi: Should this be `#listTab` to be consistent?

src/app/basket/basket.component.html
jmedinaebi: In complex-results.component.html we use `#navigatorTab` and `#list` in the `mat-tab` elements, but here we don't. Shouldn't it be the same?

src/app/complex/complex-results/complex-navigator/service/state/complex-navigator-display.service.ts
```diff
@@ -85,4 +85,56 @@ export class NavigatorStateService {
         return paramValue;
     }
   }
+
```
jmedinaebi, line 88: We should move some of the computed variables we have in different components here, to reduce duplication and also to keep the logic about the width size here.
```
isSorted = computed(() => this.componentsSorting() !== NavigatorComponentSorting.DEFAULT);  isOrganismSorting = computed(() => this.componentsSorting() === NavigatorComponentSorting.ORGANISM);
sortingWidth = computed(() => this.isSorted() ? 20 : 0);
interactorWidth = computed(() => 230 + this.sortingWidth());
```

Then we would need to update `complex-navigator.component.html`
```
<div class="ComplexNavigator" #parent
     [class.horizontal]="complexSearch().elements.length <= 6"
     [class.sorted]="state.isSorted()"
     [style.--cols]="numberOfColumns()"
     [style.--colSize.px]="state.columnWidth()"
     [style.--sizeSpaceHolder.px]="state.spaceHolderWidth()"
     [style.--intHeader.px]="state.interactorWidth()"
     [style.--sortingGroupSize.px]="state.sortingWidth()"
>
```
And `table-interactor-column.component.html`, lines 10-12
```
                  [matTooltip]="state.isOrganismSorting() ? (range.value | species) : range.value">
                <div class="interactorNameContainer" [style.--rowspanSize]="range.length">
                  @if (state.isOrganismSorting()) {
```
And line 60
```
@if (j === 0 && state.isSorted()) {
```

We can also create a new method in this class to calculate the width, as we're using mostly variables from this class already
```
  public setColumnWidth(totalComponentWidth: number, numberOfColumns): void {
    this.columnWidth.set(Math.max(
      (totalComponentWidth - this.interactorWidth() - this.spaceHolderWidth()) / numberOfColumns,
      70
    ));
  }
```

Then, in `complex-navigator.component.ts` we can simplify the `adjustColWidth` method.
```
  @HostListener('window:resize', ['$event'])
  adjustColWidth() {
    setTimeout(() => this.state.setColumnWidth(this.parent.nativeElement.clientWidth, this.numberOfColumns()));
  }
```

And finally remove the unused computed variables from the 2 components mentioned.

jmedinaebi (commented):
Just a few comments on moving some stuff to the state service class.

EliotRagueneau:
> > I've made the same calculation as you, but on the measured size of the component, and I update this measure when the window is resized, when the number of columns evolve and when the tab is selected (otherwise total is 0 if the tab was previously disselected)
> 
> I was having a look at your latest commit. I will take a better look, but it looks way more complicated than before. What's the actual visible or performance change? Before we had 1 effect, now we still that effect, but we have many other things. Is the difference now that we don't have the total width size? We calculate based on the window?

Basically you were assuming that the total size was just minColWidth * maxColNumber. With the latest commit, yes it is more complicated, but the navigator is filling the available space. e.g. when you zoom out it adjust to it

jmedinaebi:
> > > I've made the same calculation as you, but on the measured size of the component, and I update this measure when the window is resized, when the number of columns evolve and when the tab is selected (otherwise total is 0 if the tab was previously disselected)
> > 
> > 
> > I was having a look at your latest commit. I will take a better look, but it looks way more complicated than before. What's the actual visible or performance change? Before we had 1 effect, now we still that effect, but we have many other things. Is the difference now that we don't have the total width size? We calculate based on the window?
> 
> Basically you were assuming that the total size was just minColWidth * maxColNumber. With the latest commit, yes it is more complicated, but the navigator is filling the available space. e.g. when you zoom out it adjust to it

Sorry, I hadn't replied on this after my first comment. I had a better look at this and it looks good, and I understand the more complicated logic to take window resize into account. It looks ready to go, but there's no rush to merge this, that's why I haven't approved it yet.

EliotRagueneau:
If you don't mind, I'll merge it and deploy it with the QR thingy. it will actually be helpful for the screenshot on Susie's poster I think

Issue #234, opened 2024-10-09
Opening sub-groups messes the interactor order


EliotRagueneau:
e.g. with the search `pxref:("PTHR14281")`
Normal  <img width="1224" alt="Screenshot 2024-10-09 at 11 46 01" src="https://github.com/user-attachments/assets/15167c64-a96e-4881-a4e4-f34e752cfd50">

After click on Ndc80 Complex
<img width="1260" alt="Screenshot 2024-10-09 at 11 46 37" src="https://github.com/user-attachments/assets/46fd3081-b2a4-40fa-a83f-97f5d4b7ae1c">
After closing Ndc80 Complex
<img width="1230" alt="Screenshot 2024-10-09 at 11 46 47" src="https://github.com/user-attachments/assets/564720d8-8995-43ff-b7d5-e704e0f59e27">

If you look closely at the 4th column, it gains new interactors when expanding the complex, and they don't dispear upon closing. That happens I think because [CPX-5645](https://complex-portal.github.io/complex-portal-view/complex/CPX-5645) does contain those elements recursively, but somehow somethings is messed up, specially when reclosing, we don't get back to the initial state

susiehgt:
I created a branch to fix it, I think it works okay now, I have not seen any bugs created by this fix

susiehgt:
@jmedinaebi here is the bug I fixed with the branch

jmedinaebi:
@susiehgt @EliotRagueneau issue is not because of the collapse or expand. Explanation and fix here https://github.com/Complex-Portal/complex-portal-view/pull/236

The PR includes another partial fix (other part of the fix needs to happen on the back-end) for a separate issue.

I deployed it here https://complex-portal.github.io/complex-portal-view/complex/search?query=pxref:(%22PTHR14281%22)&displayMode=navigator-view

PR #236, opened 2024-10-10
Fix issues with navigator

jmedinaebi:
There are 2 issues here.

1. Issue with navigator, clustering and subcomponents.

Currently, we load the subcomponents of complexes using `susbcribe`. The logic inside the subscribe section does update the subcomponents, but not replace the objects, so the `input` and `computed` does not get triggered. So the first time we load the subcomponents we do not identify them as part of some complexes, and then they are ignored for the clustering and we don't display the stoichiometry or lines properly. When expanding and collapsing, the stoichiometry and lines is recalculated but the clustering is not, resulting in this bug https://github.com/Complex-Portal/complex-portal-view/issues/234

The fix is to create an observable with all the complexes, and after all the complexes are fully loaded (subcomponents included), then we continue (clustering, orthologs, etc.).

2. Issue with molecule sets and empty ids.

Some molecule sets have no identifiers, which cause multiple issues.

The first one is that we try to display the id, but there is none. Easy to fix, we just don't display it.
The second one is the search, which is based on ids. Easy to fix, we use the name if no identifier.
The third one is that we use the identifiers to merge components together, check if a component is part of a complex, etc.

Because of the 3rd issue, we cannot have empty identifiers. On the back-end we are gonna fix the issue, although we need to re-run the SOLR indexing. For molecule sets, if we don't have ids, we get ids from its members. On the edge case of not having any id at all (should not happen but technically possible) we then use the AC as the id. This fixes the 3rd issue, but we don't want to display or link using the AC, so we need to check the identifier is the actual component id and not the ac.


jmedinaebi:
The back-end fixed has been deployed, so it all works properly now. You can see how the molecule sets now have ids and links and they don't all end together on the same row here https://complex-portal.github.io/complex-portal-view/complex/search?query=ASC1&page=1

PR #238, opened 2024-10-11
Show loading spinner while processing the components

jmedinaebi:
When we process the complex components, we need to fetch the subcomponents of subcomplexes, and wait for that to do the clustering, so we need to wait for that to display the table. This is when people search, and also on changing pages.

I have added a loading spinner to be displayed while that is happening, so we no longer see an empty, or the table from the previous page until it gets updated.

I have deployed this, along with other changes, to the Github pages.

Worst case scenario, we have a lot of subcomplexes: https://complex-portal.github.io/complex-portal-view/complex/search?query=*&page=1&minConfidence=1&displayMode=navigator-view&mergeOrthologs=true&componentsSorting=Default&idDisplay=true&organismIconDisplay=true&typeIconDisplay=true&interactorType=stable_complex


jmedinaebi:
@EliotRagueneau I'm not sure about these changes. Let me know what you think.

Not related to the changes in this PR, but 2 things I've noticed when running Complex Portal locally.

1. When changing pages, we get this type of warning
   Screenshot 2024-10-14 at 11 30 08
   I'm calling the public back-end, so there is no duplicates, and I can also see that on the table, no duplicates. I'm guessing it's something to do with sorting. I'll keep investigating.

2. To fix the issue you reported last week, I used forkJoin to make the list of observables into an observable of a list. I already merged the changes, but the changes were the ones here https://github.com/Complex-Portal/complex-portal-view/pull/236. Edit: Actually, this is because of something else, and it also happens in other places. Still investigating.
   After those changes I'm seeing this error in the console
   Screenshot 2024-10-14 at 11 31 20

src/app/complex/complex-results/complex-navigator/table-structure/table-header/table-header.component.ts
```diff
@@ -14,12 +14,11 @@ export class TableHeaderComponent {
   complexes = input.required<Complex[]>();
   onComplexRemovedFromBasket = output<string>();
 
-  shadowVisible = input<boolean>(false);
+  shadowLeftVisible = input<boolean>(false);
```
jmedinaebi, line 17: I renamed this to make clear what shadow we mean.

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.html
```diff
@@ -35,7 +35,7 @@
             <div class="shadow right" [class.visible]="shadowVisible()"></div>
           </td>
           <!-- Interactors' stoichiometry -->
-          @for (complex of navigatorComplexes; track complex.complex.complexAC) {
+          @for (complex of navigatorComplexes; track $index) {
```
jmedinaebi, line 38: I tried multiple things, it was definitely the sorting of the complexes, but I did not manage to fix it keeping the track on complexAC, so I replaced with $index and now it's fine.

src/app/complex/complex-results/complex-navigator/table-structure/table-structure.component.ts
```diff
@@ -154,12 +165,30 @@ export class TableStructureComponent implements AfterViewInit {
     source.scrolling = true;
     target.scrollTo({left: source.scrollLeft, behavior: 'instant'});
 
-    this.setHorizontalShadowVisibility(source);
+    this.setHorizontalShadowVisibility();
+  }
+
+  @HostListener('window:resize', ['$event'])
```
jmedinaebi, line 171: We were missing the resize for this. If you resize the window and then horizontal scrolls is needed, the shadow was not appearing.

src/app/complex/complex-results/complex-navigator/table-structure/table-structure.component.ts
```diff
@@ -34,9 +43,10 @@ export class TableStructureComponent implements AfterViewInit {
   }
 
   ngAfterViewInit(): void {
-    const header = this.headerDiv.nativeElement;
-    this.setHorizontalShadowVisibility(header);
-    window.addEventListener('scroll', () => this.shadowTopVisible = header.getBoundingClientRect().top <= this.scrollStart());
+    if (!!this.headerDiv) {
+      setTimeout(() => this.setHorizontalShadowVisibility());
```
jmedinaebi, line 47: This timeout fixes the error mentioned on the comment I left the other day. I've tried multiple things but nothing seemed to work except this.

jmedinaebi:
@EliotRagueneau I managed to fixed the 2 issues I reported above. It's deployed in https://complex-portal.github.io/complex-portal-view/home

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.html
EliotRagueneau: What we should do in theory is only to trigger the "update" after the sorting is performed, like doing the sorting on a copy of the array and then replacing it. That would be better for optimisation I think

EliotRagueneau: However, not a bad idea tobe based on indexes as sometimes, we have duplicateds ACs when we have different version of the same complex it seems

jmedinaebi:
```
doing the sorting on a copy of the array and then replacing it
```

We are already doing that, and I even changed the code to make that even inside the function we call that generates the new array. Still complains.

PR #241, opened 2024-11-26
Podcast integration

jmedinaebi:
We're adding the podcast to the complex portal main page. This has been done by @susiehgt 

The MP3 file is also added in Github, although we could look into moving it to the FTP server.

What's the plan for new podcast episodes? At the moment we have one, and we display one. What happens when we have multiple episodes?


EliotRagueneau:
I would suggest to intergrate the podcast on a broader level so that they can listen to it without being stuck on thehome page for 10 minutes, like a floating player if you start it

EliotRagueneau:
ToDo:

- [x] Put the controller above the `router-outlet` so that it is available across the website if switched on, no matter in which subpage you go to, and without interruption
- [x] Display the controller in a sticky container so that people can always stop/pause it. They should also be able to close the little controller if they're tired of it,
- [x] Make the controller appear only if the user triggered it's playing from the home page, but we don't want to display the option to listen to the podcast in every single page by default (Basically we just need a button in the home page that triggers a boolean which displays the controller and launch it). Essentially the user flow should be: start it from home page, control it everywhere
- [x] Move the resource to asset folder rather than a permalink to some github repo
- [x] Make sure the mp3 meta-data is correctly annotated so that the file has a title, an author and things like that
- [x] Above point important for this one: ensure that the controller is standard enough that it gets picked up by browser features to control sound across tabs. These use meta-data to show the title and author of the audiobeing played, even some pictures  (see difference bellow between properly annotated or not)
Screenshot 2024-12-02 at 17 20 30

EliotRagueneau:
For the workflow, you can have a look at the guided toor of reactome GSA, we probably want something relatively similar in the way of "start it from home page, control it everywhere" https://reactome.org/gsa/home

EliotRagueneau:
@susiehgt You can have a look at how I did things now if you want to.
@jmedinaebi would like a little review from you to check you're happy with it too :D

EliotRagueneau:
Screenshot 2024-12-03 at 16 58 56
The chrome audio aggregator does work well with the image now as well

src/app/home/podcast/podcast.component.html
```diff
@@ -0,0 +1,30 @@
+@if (podcast.isOpen()) {
+  <div class="floating-container" cdkDrag cdkDragBoundary="body">
+    <span class="title">
+      @if (podcast.episodes()?.length > 1) {
+        <button mat-icon-button (click)="podcast.previous()" [disabled]="podcast.isFirst()"
+                *ngIf="podcast.episodes()?.length > 1">
```
jmedinaebi: This `*ngIf` is no longer neede,d right? We have the `@if` just above.

src/app/home/podcast/podcast.component.html
```diff
@@ -0,0 +1,30 @@
...
+
+      <div class="auto-scroll">
+          <div class="gradient-overlay"></div>
+          <div class="content">
+            {{ podcast.currentEpisode().title }}
+          </div>
+      </div>
+      @if (podcast.episodes()?.length > 1) {
+        <button mat-icon-button (click)="podcast.next()" [disabled]="podcast.isLast()"
+                *ngIf="podcast.episodes()?.length > 1">
```
jmedinaebi: Same as previous comment.

src/app/home/podcast/podcast.component.scss
```diff
@@ -0,0 +1,65 @@
+.floating-container {
+  position: fixed;
+  bottom: 0.7rem;
+  left: 0.7rem;
+  background: rgba(255, 255, 255, 0.54);
+  border-radius: 25px;
+  backdrop-filter: blur(4px);
+  width: 16.5%;
+  padding: 6px 6px 0 6px;
+  box-shadow: 0px 4px 4px #cdcdcd;
```
jmedinaebi: Should we update the shadow to `box-shadow: 4px 4px 4px 4px #cdcdcd;`?

As it currently is
Screenshot 2024-12-04 at 10 27 48

After the change
Screenshot 2024-12-04 at 10 27 35

src/app/home/podcast/podcast.service.ts
```diff
@@ -0,0 +1,84 @@
+import {computed, effect, Injectable, model, Signal, signal, WritableSignal} from '@angular/core';
+import {HttpClient} from '@angular/common/http';
+// import {Reader} from 'jsmediatags';
```
jmedinaebi: Delete

src/app/home/podcast/podcast.service.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  private listEpisodes(): Observable<AudioFile[]> {
+    return this.http.get<Partial<AudioFile>[]>('assets/audio/audio-manifest.json').pipe(
+      map(audioFiles => audioFiles.map(file => ({...defaultFile, ...file, url: `assets/audio/${file.file}`} as AudioFile))),
+      tap(audioFiles => console.log(audioFiles))
+    );
+  }
+
+  // private loadMetadata(partial: Partial<AudioFile>): Observable<AudioFile> {
```
jmedinaebi: Delete

jmedinaebi (commented):
Looking really good! I just left a few small comments

src/app/home/podcast/podcast.component.scss
EliotRagueneau: I tend to prefer the first one as it's a bit more subtle, but I guess that's a matter of taste. @susiehgt you have an opinion on the matter?

EliotRagueneau: Screenshot 2024-12-05 at 09 47 18
I improved it a tiny bit with an internal shadow so that we can distinsuish the whole player, if that's what was bothering you @jmedinaebi

EliotRagueneau (commented):
Great now

PR #243, opened 2024-12-03
Feature/bluesky embedded

EliotRagueneau:
Use bluesky instead of twitter in order to support embedding.
Style can probably be improved, but it's already not too bad.

Not a big fan of the style injection as a string though, might be good to find an alternative that Webstorm support for autocompletion, but it's not so easy :/


jmedinaebi:
It's looking good. I tried to use it the other day, but gave up after a while because of the styles. 

A few comments:

1. Should we keep the follow us on X button? Have the section named social media and have both widgets.
   Screenshot 2024-12-04 at 09 49 18

2. The widget is wider than the side panel where it is. It's not a big issue, but it would be nice if it all was the same size
   Screenshot 2024-12-04 at 09 52 06 (2)

3. Can we add some shadow to the widget? At the moment the widget background is white and the complex portal background is white, so the text is just floating there. With a border, or shadow, it will stand out more.

jmedinaebi:
Making the home page expanded, with some padding, and adding back the Twitter widget, it would look something like this

Screenshot 2024-12-04 at 10 19 22

I'm not sure about these changes. But if we go with expanding the home page, we need to check nothing else looks weird.

EliotRagueneau:
> It's looking good. I tried to use it the other day, but gave up after a while because of the styles.
> 
> A few comments:
> 
> 1. Should we keep the follow us on X button? Have the section named social media and have both widgets.

Not a bad idea

>    Screenshot 2024-12-04 at 09 49 18
> 2. The widget is wider than the side panel where it is. It's not a big issue, but it would be nice if it all was the same size

That's how it was orginally, but tbh it looked worse because the skeets text hd not enough space, so they were super tall. What we have can probably be improved though
>    Screenshot 2024-12-04 at 09 52 06 (2)
> 3. Can we add some shadow to the widget? At the moment the widget background is white and the complex portal background is white, so the text is just floating there. With a border, or shadow, it will stand out more.

Yes we can do more or less what we want with it: the only annoying part is that we have to put the styling in a string attribute, but we can quite easily customise it.

jmedinaebi:
I managed to get the widget thinner, although it looks a bit weird when it is too thin, by changing the style to
```
    <bsky-embed
      username="complexportal.bsky.social"
      custom-styles="
        img[alt='profile picture'] {
            display: none;
        }

        section {
            align-items: baseline;
            max-height: 406px;
            overflow-y: overlay;
            overflow-x: clip;
            width: min-content;
            box-shadow: 0 0 5px 0 rgba(0, 0, 0, 0.3);
        }

        .p-4 {
            padding: 0.75rem;
        }

        .max-w-screen-sm {
            max-width: 280px;
        }
  "></bsky-embed>
```

Changes on the style are:
- max-height updated as twitter widget was added back
- `[id^='post-']` width property removed
- `overflow-x: clip;` added in section
- `.p-4` padding changed to add padding on the left, but smaller padding overall
- max-width set in `.max-w-screen-sm` to control the width of the widget. This is the property we can play with to change the width.

After that, adding the twitter widget back, and adding the green bottom border on the Social Media title, it looks like

Screenshot 2024-12-04 at 10 46 30
Screenshot 2024-12-04 at 10 46 34

PR #244, opened 2024-12-13
feature/improve-navigator-ordering

EliotRagueneau:
Fix 2 problems:
- Avoid floating predicted complexes components because they are "not found" in any complex (Ndc80 search) 
- Avoid instability in results arising from race conditions between timesAppearing definition and timeAppearing usage

Also improve sorting of components when similar timesAppearing and similar first index appearing.

This improvement needs to be discussed and reviewed because it is significantly longer (+17,52%  compute time for a relatively big example (20 complexes x 272 components, still marginal time though from 7ms to 8,5ms)

See comparison at https://docs.google.com/spreadsheets/d/1oG9kVww2F1s5U2EjoataKmOiVaY4yx4dXloqJJGqBAc/edit?usp=sharing


EliotRagueneau:
Previous sorting: Sort by firstIndex, then if equal, by amount appearing
Screenshot 2024-12-13 at 17 19 37

New sorting: Sort by all indexes of apparition
- if component A appears more time than B, then component A is placed first (keeping the behaviour of gathering shared component at the end for the stairs effect)
- If they have the same first index of apparition, we compare the second, and then the next ones until there is a difference or we ran over all apparitions. This makes basically sorted first by the first complex they appear in, then by the sceond complex they appear in, etc, creating again a kind of staircase effect
Screenshot 2024-12-13 at 17 21 27

EliotRagueneau:
These are extremely rare conditions though, as you need to have different complexes having the same amount of component shared with another one, but not the same components. Here for instance all components have a first index of 0, so then they are all compared based on the timeAppearing, but they are all appearing only 2 times, so they are considered equal by our sorting algorithm, and therefore don't get sorted between one another.

Therefore, I think the proposed solution is better, but I don't know if it's worth a 20% increase of compute time, what do you think @jmedinaebi ?

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
```diff
@@ -26,6 +23,8 @@ export class TableInteractorColumnComponent {
 
   isOrganismSorting = computed(() => this.state.componentsSorting() === NavigatorComponentSorting.ORGANISM);
 
+  navigatorComponentsTA = computed(() => this.calculateTimesAppearing(this.complexes(), this.navigatorComponents()));
```
jmedinaebi: What's TA?

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
```diff
@@ -206,9 +204,12 @@ export class TableInteractorColumnComponent {
   }
 
   private compareFn(a: INavigatorComponent, b: INavigatorComponent) {
-    return -(b.indexAppearing - a.indexAppearing) || // First by order of appearance (staircase effect)
-      -(b.timesAppearing - a.timesAppearing); // Then by reversed occurrence, in order to minimize edge length
-  }
+    for (let i = 0; i < Math.max(a.indexAppearing.length, b.indexAppearing.length); i++) {
```
jmedinaebi, line 214: Logic is a bit hard to follow, can you leave the comments in?

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
```diff
@@ -292,6 +289,7 @@ export class TableInteractorColumnComponent {
         }
       }
     }
+    return navigatorComponents;
```
jmedinaebi: In this method, we are updating `this.navigatorComponents()` and also returning and setting the returned value to `navigatorComponentsTA()`, so `this.navigatorComponents()` and `navigatorComponentsTA()` are equals. We should treat `this.navigatorComponents()` as immutable, creating a new array and returning the new array.

jmedinaebi (commented):
Looks good, just a couple of minor comments.

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
EliotRagueneau: Times Appearing, but it was making things too long for the linter.
But I agree it's not ideal

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
EliotRagueneau: Yeah good point

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
jmedinaebi: Also, are we not using `navigatorComponentsTA` in the HTML at all?

jmedinaebi: I just pushed a change for renaming `navigatorComponentsTA` and also to use the new variable instead of `navigatorComponents`.

jmedinaebi:
I have pushed a new commit to show the loading spinner when the user makes a new query, instead of just showing the old results until the new ones come in. I have deployed my changes (this and the previous commit) to the github pages.

I have also restarted Complex-WS and SOLR in the public instances to see if that improves things, as queries were being slow this morning.

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
```diff
@@ -23,7 +23,7 @@ export class TableInteractorColumnComponent {
 
   isOrganismSorting = computed(() => this.state.componentsSorting() === NavigatorComponentSorting.ORGANISM);
 
-  navigatorComponentsTA = computed(() => this.calculateTimesAppearing(this.complexes(), this.navigatorComponents()));
+  sortedNavigatorComponents = computed(() => this.calculateTimesAppearing(this.complexes(), this.navigatorComponents()));
```
EliotRagueneau: They are not really sorted in this, just preprocesed to be ready to sort . Could be sortingReadyComponents

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
```diff
@@ -34,7 +34,7 @@ export class TableInteractorColumnComponent {
   };
 
   constructor(public state: NavigatorStateService) {
-    effect(() => this.sortNavigatorComponents(this.navigatorComponentsTA(), this.complexes(), this.state.componentsSorting()));
+    effect(() => this.sortNavigatorComponents(this.sortedNavigatorComponents(), this.complexes(), this.state.componentsSorting()));
```
EliotRagueneau: the sorting is actually only taking place here

EliotRagueneau:
#d388b28 Is really important because I realised every time we update the display options or the sorting or grouping, we were triggering a new search API call, which was making everything slow. Thanks to this one, we now send requests only when an actual parameter of the search has changed. Also make sures that we only triggers the preparation of components and the sorting when actually needed, not whenever you change something.

src/app/app.component.ts
```diff
@@ -53,7 +53,7 @@ export class AppComponent implements OnInit, AfterViewInit {
         ProgressBarComponent.display();
         return;
       }
-      window.scrollTo(0, 0);
+      // window.scrollTo(0, 0);
```
jmedinaebi, line 56: What was this doing? And why are we removing it now?

src/app/complex/complex-results/complex-navigator/table-structure/model/navigator-component.model.ts
```diff
@@ -28,23 +28,29 @@ export interface INavigatorComponent {
...
+
+  set timesAppearing(value: number);
+
+  get indexAppearing(): number[];
+
+  set indexAppearing(value: number[]);
+}
+
+export function isSortable(component: INavigatorComponent): component is SortableNavigatorComponent {
+  return (component as SortableNavigatorComponent).timesAppearing !== undefined;
```
jmedinaebi, line 55: Why not `!!(component as SortableNavigatorComponent).timesAppearing`?

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
```diff
@@ -20,13 +17,19 @@ interface Range {
...
-  navigatorComponents = input<INavigatorComponent[]>();
+  complexes = input.required<Complex[]>();
+  navigatorComponents = input.required<INavigatorComponent[]>();
   shadowVisible = input<boolean>(false);
 
   isOrganismSorting = computed(() => this.state.componentsSorting() === NavigatorComponentSorting.ORGANISM);
 
-  navigatorComplexes: NavigatorComplex[];
+  sortableComponents = computed(() => this.prepareSorting(this.complexes(), this.navigatorComponents()));
+  sortedComponents = computed(() => this.sortNavigatorComponents(
```
jmedinaebi: `sortableComponents` and `sortedComponents` are the same. They are just 2 references to the same array. We call `sortNavigatorComponents`, which ends up doing `sortableComponents.sort(...)` and return that, so the two arrays are the same one. It'd be better to have just one to not create confusion on what's the difference and which one to use.

src/app/complex/complex-results/complex-results.component.ts
```diff
@@ -177,19 +191,51 @@ export class ComplexResultsComponent implements OnInit, AfterViewInit {
...
+ * Used to get the new object reference if its data is different from the initial one, or the initial one. <br>
+ * Allows Signals update (based on object ref) to only get triggered if there is an actual data change. <br>
+ * The difference is determined by going through the whole object recursively, and by using only value comparison, no reference comparison
+ *
+ * @param initialObject The initial value of the object,
+ * will be returned if <b>no data differs</b> from potentialUpdatedObject
+ * @param potentialUpdatedObject The potentially updated version of the initialObject,
+ * will be returned if <b>data differs</b> from initialObject
+ */
+function keepObjectIfSimilar<T>(initialObject: T, potentialUpdatedObject: T): { object: T, similar: boolean } {
```
jmedinaebi: Should it be called `keepObjectIfSame` instead of similar?

jmedinaebi (commented):
Just a few comments on the latest changes.

src/app/app.component.ts
EliotRagueneau: It was scrolling to the top of the page every time the URL was updated, but since we update the URL everytime we change a filter or an option or things like that, it's actually more annoying than anything.

src/app/complex/complex-results/complex-navigator/table-structure/model/navigator-component.model.ts
EliotRagueneau: Might create problem with falsy values like 0

src/app/complex/complex-results/complex-navigator/table-structure/table-interactor-column/table-interactor-column.component.ts
EliotRagueneau: Yeah I know they are the same, it was to make more sense semantically, but indeed we could remove the sorted one

src/app/complex/complex-results/complex-results.component.ts
EliotRagueneau: The only problem with Same is that it could be understood as same reference. Maybe `keepObjectIfSameProps`?

src/app/complex/complex-results/complex-navigator/table-structure/model/navigator-component.model.ts
jmedinaebi: Yeah, `timesAppearing` being a number or an array creates issues with `!!`. As long we're sure it cannot be null, just undefined or an actual value, this should be fine then.

jmedinaebi (approved):

PR #249, opened 2025-05-12
Display stoichiometry range

jmedinaebi:
Currently, we only display the min stoichiometry in the list of participants in the complex details page. This is a small change to display a range instead of min and max are different.


jmedinaebi:
Currently, the blue circle where the stoichiometry is displayed, looks good with a single number, but with a range it looks a bit weird
Screenshot 2025-05-12 at 11 16 08

I have increased the size of that circle to display the range properly, but I have increase it all the time
Screenshot 2025-05-12 at 11 19 01

Maybe it's too much. @EliotRagueneau thoughts on this?

EliotRagueneau:
Increasing the size is not too bad, but for high stoichiometry, it will become a problem any. I think a better idea is to make it become a pill shape isntead of a circle if the label is more than one character. ie. `corner-radius:999px` should worrk like a charm I think

jmedinaebi:
It looks much better after using `border-radius`.

Screenshot 2025-05-19 at 09 52 52

jmedinaebi:
I have also updated the complex viewer version and now we have ranges displayed in the viewer and the legend
Screenshot 2025-05-22 at 14 11 31

PR #252, opened 2025-06-30
Display IEDB xrefs

jmedinaebi:
New section to display xrefs of protein and other components of the complexes. We don't want to display all protein xrefs because we are not Uniprot, so we filter which specific xrefs we want to display. At the moment we only want to display IEDB xrefs.

Currently deployed in the Github pages, check https://complex-portal.github.io/complex-portal-view/complex/CPX-12142#components-references and https://complex-portal.github.io/complex-portal-view/complex/CPX-18933#components-references


src/app/complex/shared/service/complex-portal.service.ts
```diff
@@ -47,6 +48,16 @@ export class ComplexPortalService {
     );
   }
 
+  getComplexComponentsXrefs(ac: string, databases: string[]): Observable<ComplexComponentDetails[]> {
```
jmedinaebi, line 51: This depends on changes on the backend for a new endpoint that have not been deployed in the production webservice yet.

PR #253, opened 2025-06-30
Display features

jmedinaebi:
Display complex components features, including InterPro xrefs right under the complex viewer.

Currently deployed in the Github pages, check https://complex-portal.github.io/complex-portal-view/complex/CPX-310 and https://complex-portal.github.io/complex-portal-view/complex/CPX-9206


src/app/complex/shared/model/complex-details/features.model.ts
```diff
@@ -1,10 +1,21 @@
+import {CrossReference} from './cross-reference.model';
+
 /**
  * Created by mkoch on 19/07/2016.
  */
 export interface Features {
```
jmedinaebi, line 6: This depends on changes on the backend to return additional fields for the features.

src/app/complex/complex-details/complex-participants/complex-participant-features/complex-participant-features.component.html
```diff
@@ -0,0 +1,122 @@
...
+          <i class="icon icon-common"
+             data-icon="&#xf078;"
+             style="float: right"
+          ></i>
+        </h4>
+      </a>
+      @if (featuresExpanded) {
+        <table class="hover small-font">
+          <thead>
+            <th style="width: 12%">Name</th>
```
EliotRagueneau: not sure width:12% is really the best way to size the columns. Maybe making the parent flex and the one that is suppoed to be bigger with `flex: 2 2 auto` while the others with `flex: 1 1 auto`.  And let's put that in the css to avoid redundancy by making something along those lines
```css
tr {
 display: flex;
}
tr > th ,
tr > td {
 flex: 1 1 auto;
}

.flex-2 {
 flex: 2 2 auto;
}
```

also needs a `tr` wrapping `th`s

src/app/complex/complex-details/complex-participants/complex-participant-features/complex-participant-featuress.component.ts
EliotRagueneau, line 1: remove last s of file name

EliotRagueneau (changes requested):
Looks okay to me, just change the way you size the columns to something more future proof (e.g. if we add or remove columns)

src/app/complex/complex-details/complex-participants/complex-participant-features/complex-participant-features.component.html
jmedinaebi: I changed this to use flex with the changes in the CSS you suggested and it looks like this
Screenshot 2025-07-17 at 11 51 07

Each row is applying flex differently, so the table does not look like a table.

If I remove the styles and don't add the changes to the CSS file, it looks like this
Screenshot 2025-07-17 at 11 58 37
Screenshot 2025-07-17 at 11 58 43
Screenshot 2025-07-17 at 11 58 49

And with the styles to set the columns to 12 or 16%, it looks like
Screenshot 2025-07-17 at 12 00 15
Screenshot 2025-07-17 at 12 00 22
Screenshot 2025-07-17 at 12 00 29

On these examples, nothing is too long to not be displayed properly, I'm gonna try to find an example like that to see what happens.

jmedinaebi: Still looks ok either way.

With the styles and the width %
Screenshot 2025-07-17 at 12 06 33

Without the styles
Screenshot 2025-07-17 at 12 06 44

I think the only relevant thing I was trying was to have more space for the ranges, but it's not that important. I can remove the styles and leave it how it looks by default.

EliotRagueneau: Sorry for that, somehow I believed this would work, but I was completely wrong ^^' 
I do agree that is is good to have longer ranges, yet not essential.

While trying to understand why what I proposed doesn't work, I discovered a new tag which could be quite useful in that context though, at least to avoid repetition of style which is prompt to errors: `colgroup` https://www.w3schools.com/tags/tag_colgroup.asp

I think if we used it, I would be more fine with using styles on just the first row let's say.
Another approach I saw was to use `table-layout: fixed` which ensures same width for each column. We can then have double width for one by having each of its cell at `colspan=2`. This has some slight issues as we need to then treat what happens in case of overflow, but might be worth trying as it is quite easily maintainable. the only annoying bit being to need to specify `colspan="2"` on each row and not just one.

Tbh this is really a point of detail and like not that big of a deal, so I'm also fine leaving it there to the table without styles, I let you decide what you prefer among all options. I'll approve whichever you prefer

jmedinaebi: I removed the styles and left it as default, with columns adjusting the width to the content.

EliotRagueneau (approved):

PR #254, opened 2025-10-13
Display release dates after the complex header

jmedinaebi:
Display first and latest release dates for a complex. If they are the same date, then we display it just once.

Draft as we need to wait until the next release, and also, the backend changes to be deployed.


jmedinaebi:
Screenshot 2025-10-07 at 10 11 47
Screenshot 2025-10-07 at 10 11 52
Screenshot 2025-10-07 at 10 33 41

